PR #557, opened 2021-02-10
Delete .DS_Store

mane:
This is a hidden file created by MacOS which shouldn't be committed into git.


Luis-Fernando-Molina:
Thanks for spotting this!

PR #558, opened 2021-02-11
(typo) repo url fix

kazazes:


Luis-Fernando-Molina:
Hey Peter, the URL is correctly pointing to the user's fork. We have an in-app command that facilitates documentation contribution by automating the creation of pull requests which requires users to have their own fork of the repo in order to work. So in this case, the URL is correct since they have to clone their fork, not the main repo in order for this to work.

PR #609, opened 2021-03-03
Contribution: Fixed bug when logging in getOHLCVs() and flat array was incorrectly being addressed as a JSON object.  Also a couple of spelling mistakes.

suttridge:
Can we ditch this if it is wrong - I wanted to do the pull to the develop branch, not the master branch.


suttridge:
This was all wrong - it was supposed to go to the develop branch, and not to the master branch.

PR #633, opened 2021-03-07
WIP: Feature/docker

unfernandito:
## Docker is now part of Superalgos

- Added Dockerfile with LTS Nodejs version and noBrowser run by default
- Added dockerignore
- Added github action to automatize new versions


.github/workflows/docker.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to GitHub Packages
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          registry: docker.pkg.github.com
+          repository: unfernandito/Superalgos/superalgos
```
unfernandito, line 18: The name of repository should be provide by the repo itself

PR #635, opened 2021-03-07
Contribution: Minor Doc Corrections

nickspurry:


Luis-Fernando-Molina:
Hey! thank you for sending these corrections, but I can not merge them directly on master. You should send the PR to the develop branch instead. The app facilitates this by allowing to select the branch at the footer of the docs. You can use the commands:

app.update
then switch to develop branch
app.contribute

cheers!

PR #700, opened 2021-03-18
Bug Fix: Telegram Announcement Formula and Condition Changes

Smidy13:
Bug Fix for Telegram Announcements. This now ensures the JSON data received by the Announcement.js contains both the announcementFormula and announcementCondition (TradingSessionFunctions.js changes). Then the makeAnnoucement function has been changes to properly validate the announcementCondition and use the announcementFormula if valid.

(side note, this is my first Pull Request, sorry for any issues!)


Luis-Fernando-Molina:
Awasome! Thanks for this fix. 2 things:

1. I removed a console.log
2.                         const message = 'Announcement Formula Evaluation Error'  you need to create at the Docs an extra topic page for the topic TS LF Trading Bot Error, so that when this error happens and the user at the UI clicks on the error, then the docs will open at this new page you added. 

I am merging this anyways, then you can submit the new docs page.

PR #741, opened 2021-03-30
Add files via upload

Basalt09:


julian-molina:
These seem to be old commits. I can't merge because they come from the master branch. Still, I think those icons have been submitted already. Will let you know if any of those is missing...

PR #773, opened 2021-04-05
Bug Fix for Announcements + Browser Type / Version Check

Smidy13:
This bug fix moves the lightingPath entries for Announcements down further so that they are after the Create Order and Cancel Order event (and therefore get correctly added to the Create Order and Cancel Order Event).

It also adds a entry for making an Announcement as we move into the Manage Stage (as there was none but you can create Announcements for the Manage Stage).


Smidy13:
Added in a browser check based on a support query in the Support Telegram group. I'll leave it up to you if you choose to merge this particular commit or not. I figure you might want to leave the browser check out (as others may be using other browsers and this could put them off). I just figured we should add something as if a user is using a lower version of Chrome it won't load with no visible warning or error to the end-user.

Luis-Fernando-Molina:
good job!

PR #780, opened 2021-04-07
Bug Fix: Getting Started Tutorial Not Loading

Smidy13:
At this stage just submitting a quick bug fix so the Getting Started tutorial loads again.

Will add to my TODO list is investigate a better way of checking that the initial indexing has been done and the design space is loaded before the playTutorial function is called. But this will require a bit of a code dive to understand the sequence.

### DIFF
--- a/Projects/Superalgos/UI/Spaces/Tutorial-Space/TutorialSpace.js
+++ b/Projects/Superalgos/UI/Spaces/Tutorial-Space/TutorialSpace.js
@@ -892,7 +892,6 @@ function newSuperalgosTutorialSpace() {
     }
 
     function playTutorial(node) {
-        if (UI.projects.superalgos.spaces.designSpace.workspace.isInitialized !== true ) { return }
 
         PAGE_NUMBER = 0
         TUTORIAL_NAME = node.name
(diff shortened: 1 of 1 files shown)


Luis-Fernando-Molina:
These IFs you are removing are there as a safeguard, since if the workspace is not yet initialized the tutorial will not be loaded anyway as at workspace is not initialized and it might crash in a more unexpected way. At least with the Ifs it will only not show the tutorials. Maybe we should wait until you find the root cause of why this functions are called before the workspace is initialized before merging this PR?

The workspace should be initialized after the indexing of the docs is finished, but obviously that is not happening always. See if you can find out why sometimes these functions are called before that please.

NOTE: it might be also true that removing the IFs make it work, I haven't tried it. Maybe the workspace initialization is advanced enough not to crash, but I don't know.

Smidy13:
Sounds good and completely agree! I'll continue investigation tomorrow and work on a permanent fix.

My concern was this bug may be affecting new user adoption due to the tutorial not loading thus wanting to get in a quick fix.

backamblock:
i waited for 30 min now and it doesn't come up after first install on a powerful machine

PR #804, opened 2021-04-12
WIP: attempt to close the trade when not found

ssplatt:
work in progress

When cancelling an order at Coinbase, the following `getOrder` will result in `NotFound` because the trade can no longer be accessed. This also means that the `order` object will not have any of the order data, so "synchronization" is not possible.

state of the code:
- The order can be placed
- The order can be cancelled
- `order.status = 'NotFound'` is returned on the following `getOrder`
- internal accounting is attempted to sync the state with the exchange, but the `order` object has no accounting data, resulting in errors when `syncWithExchange()` is executed. If this function is skipped, the next orders attempt to resize to `0` which results in no order being placed.


Luis-Fernando-Molina:
Will check this after the weekend, since we are getting ready to switch branches for documentation and plugins in these days.

Luis-Fernando-Molina:
@Smidy13 can you help @ssplatt with this please?

Luis-Fernando-Molina:
This PR is getting old... can you please tell me where are you with this? Should I close it?

ssplatt:
I haven't had time to test recently. probably best to close for now and reopen when progress is made.

PR #811, opened 2021-04-15
Update README.md

DB-Sektor:
Added note to point 4. to clarify that this step is only for already Beta 8 users. That wasn't very clear for me when I was installing the software :)

## 4. Migrating from Superalgos Beta 8 (skip this step if you are a new user)


julian-molina:
Thanks for the submission. Unfortunately, we can't merge changes to master directly. Please submit to the develop branch instead. If you submit after Sunday 18th, then it must be submitted to the new plugins-docs branch. Also, it's best to add a note below the title than to modify the title itself.

PR #836, opened 2021-04-24
Create Windows Shortcut Script

harrellbm:
Superalgos-shortcut-windows

Also grouped everything under GUI-Launch folder to keep these launch scripts from cluttering up the main directory.


Luis-Fernando-Molina:
The folder GUI-Launch is not the best name, since you are launching the Client and the GUI at the same time. Can you please rename it to Launch-Scripts?

harrellbm:
That makes sense. I'll do that right now!

Luis-Fernando-Molina:
Please don't forget to update the docs / readme to help users with these launch scripts.

harrellbm:
You got it. 👌🏼 It's on my tasks list.

PR #888, opened 2021-05-08
Contribution: Grammatical and instructional fixes

kizniche:


kizniche:
I found I needed to double-click the detach icon for it to actually detach. I tried this several times, and all would only detach when double-clicking, not single-clicking.
Also, the word "Open" was a link/tooltip to an irrelevant term for what the word meant in this context. Changing to a different word with a similar meaning fixed this issue.
Last, a simple change from plural to singular, because it's referencing a single node rather than than multiple nodes.

julian-molina:
@kizniche thanks for the fixes! Can you please submit them to the `plugins-docs` branch instead? We are using that branch for updates that don't affect the core code base so that they may be merged into `master` more often...

kizniche:
Strange. I was on the plugins-docs branches when I submitted these through the UI. This is probably another bug that goes along with the issue I created for this PR.

kizniche:
Okay, now I'm not so sure. I referenced the steps I wrote in the issue and I wrote switch to Develop brance, but I'm pretty sure I was on plugins-docs as stated in the docs, and that was an error on my part writing the steps. In any case, I'll try to reproduce my issue from scratch and see if the PR goes to the plugins-docs branch. Sorry for the inconvenience.

julian-molina:
Make sure the Contributions Branch is also set to Plugins-Docs... Thanks for your efforts!

PR #892, opened 2021-05-08
Contribution: Grammar, instruction, and improper linking fixes

kizniche:


kizniche:
Now on the proper branch this time!

I found I needed to double-click the detach icon for it to actually detach. I tried this several times, and all would only detach when double-clicking, not single-clicking.
Also, the word "Open" was a link/tooltip to an incorrect term in this context. Changing to a different word with a similar meaning removed the link.
Last, a word change from plural to singular, because it's referencing a single node rather than multiple nodes.

PR #900, opened 2021-05-09

superharoun:


superharoun:
Don't get why it goes to master and not plugins-docs

PR #902, opened 2021-05-09
Fix typos

superharoun:


julian-molina:
@superharoun you made it! Thanks for your first contribution!

PR #907, opened 2021-05-11
Deleted node_modules

trasherdk:
Pushing `node_modules` to GitHub is a bad idea.

Executing `$ ( cd Client; npm ci ); node run.js` fixes that.


Luis-Fernando-Molina:
Initially we did it like this to minimize support questions. It is overdue to fix it.

After your changes, does it change the procedure to install or run the software?

Also: there are more node_modeles in inside directories, for some tasks that run from there. Did you delete those too? Asking because the huge number of files deleted does not allow me to check this clearly.

It would be a good idea to remove it from everywhere, otherwise it would be confusing for users since for instance CCXT is a dependency at 2 different levels and might run more than one version at the time.

trasherdk:
Removing `node_modules` require you to run `npm ci` in any directory that contain a `package.json`.
The `ci` is to prevent `npm` from making changes to `package-lock.json` durring installation.

I did not dive into any sub-directories, but noticed the initial clone took *a while* to complete on a 1Gb connection.

Seeing the number of `web3` security warnings made me step back, deciding to come back once those have been addressed.

harrellbm:
If we want to move forward with this right now, I could help write install scripts to run node ci in all the necessary directories.

Luis-Fernando-Molina:
@trasherdk Thanks for the clarification. @harrellbm yes, please do it so we move forward and leave these node_modules behind.

harrellbm:
@Luis-Fernando-Molina Sounds good, I will need some help with the script for Mac, but it shouldn't be terrible to port over once I finish the Linux script.

Luis-Fernando-Molina:
this PR was replaced by https://github.com/Superalgos/Superalgos/pull/950

PR #954, opened 2021-05-21
Contribution: Updated tested exchanges to indicate that Luno exchange is not longer suitable for use.

suttridge:


Luis-Fernando-Molina:
Welcome back!

PR #1023, opened 2021-05-29
English Language Fixes

SunPowered:
Some language fixes I found while introducing myself to the SA ecosystem.  Most of the fixes are minor grammatical changes.

The most drastic edit, IMO, was the renaming of some files in the Basic Education Tutorial, due to typos in the title/filename.  This broke the tutorial flow unless I updated the respective tutorial node config in the workspace to the updated title.  For this to be permanent, the 'Getting Started' Plugin Workspace was overwritten.  If there is a less extreme way to modify the Tutorial filenames, I would be happy to make that fix instead, but it might be the best way in this case.


julian-molina:
Hey @SunPowered, thanks for this contribution!

At this point, the Getting Started Tutorials workspace is undergoing a significant upgrade preparing the release of Beta 10, so I can't merge the PR in it's current form. 

Can you please remove the corresponding commit from the PR and submit the rest of the pages in a separate one? I will take care of fixing the tutorial config myself to fit the new name of the page.

SunPowered:
@julian-molina, I appreciate the feedback.

No problem, I will split the PR up to avoid touching the Getting Started Workspace directly

PR #1035, opened 2021-05-31
Bug Fix: Tutorial Reference Missing

Smidy13:
Found that the project variable is sometimes set to "Ethereum" in rare cases if you change Workspaces prior to refreshing the browser. Changed the code so it pulls from the nodeConfig.docs.project rather than the project variable. This makes more sense than using the global project variable (as this is used for loop activity loading all Schema documents where here we just want one).


Luis-Fernando-Molina:
wow!!! you made it!!! congrats!!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #1046, opened 2021-05-31
Contribution: Added Omega Data Mine with reference to Masters trading mine ; Added 2 new supertrend with 9-2 period and factor and 14-3 perdiode + factor to prepare triple supertrend strategy

Raplh2502:


9808us:
@julian-molina we can close this PR
I got in touch with the user and there are missing references that prevent the indicators to work and install correctly

PR #1132, opened 2021-06-12
Update Masters-Lightning-Strike.json

feazby:
To help fix tutorial bug
https://media.discordapp.net/attachments/836560777325445141/852733392927195156/Screen_Shot_2021-06-10_at_6.54.20_PM.png?width=2156&height=1259


Luis-Fernando-Molina:
Hey! can you please elaborate which bug this is fixing?

Luis-Fernando-Molina:
As per the image, the it mentions a reference from the Trading Session to the Trading System. That can not be fixed with the Trading System itself. Can you explain please  what was the problem and the solution?

Luis-Fernando-Molina:
Ok, looking at the issue I see that this might fix it because the Trading System was passed to master with some refactoring only valid at the develop branch... How did you fix the Trading System?

Luis-Fernando-Molina:
The issue spanned also the Trading Mine, I fix them both at my local. Thanks for submitting this and raising the alarm.

PR #1145, opened 2021-06-15
Plugins docs

quantum8:
Some minor spelling and grammar fixes


julian-molina:
Hey @quantum8... thanks for these fixes. Some of these commits involve modifying the the `type` field, that is, the title of the page. The thing is that modifying the title requires two additional jobs:

1. Adjusting the name of the file accordingly.

For instance, `trading-bot-layers-015-tutorial-step-elefant-what-elefant.json` would need to change to `trading-bot-layers-015-tutorial-step-elephant-what-elephant.json`

2. Adjusting the `type` field under the `docs` section of the configuration of the tutorial step calling the affected page.

This last point requires modifying the workspace, making sure the workspace is left in the exact same condition, so that it work as expected. This may require to RESET the tutorial after you test the changes, and may require to hit PLAY in the tutorial before you save the workspace, so that the tutorial starts automatically once the plugin workspace is loaded.

Not sure if you're familiar with all these. If you are, then you may go ahead an do it. If not, you may want to remove the commits that involve this kind of work.

Let me know!

quantum8:
Thanks @julian-molina makes sense. I guess a find/replace on those words should fix everything relating to that rather than just the initial spelling fix I've made. I'll give it a go!

julian-molina:
@quantum8 I believe the changes in the plugin workspace are pending, right? I'll wait for that before merging all this, so as not to break anything... Thanks!

quantum8:
@julian-molina I think they should be good to go now unless I've missed anything! New to all this but want to help out in any way I can!

julian-molina:
@quantum8 you haven't submitted the workspace file, so I guess you are missing the second point I mentioned earlier:

2. Adjusting the `type` field under the `docs` section of the configuration of the tutorial step calling the affected page (the ones in which you changed the titles (`type` field in the file).

This last point requires modifying the workspace, making sure the workspace is left in the exact same condition, so that it work as expected. This may require to RESET the tutorial after you test the changes, and may require to hit PLAY in the tutorial before you save the workspace, so that the tutorial starts automatically once the plugin workspace is loaded.

PR #1146, opened 2021-06-15
Update package-lock.json

wyattmcgehee:
Bump version of https://www.npmjs.com/package/ccxt
May need to update this dependency in other sub-projects soon.


Luis-Fernando-Molina:
Hey! thanks for the PR and finding and testing this. There is one more place where CCXT dependencies are needed, can you please also fix that  package-lock too? The second place is used for live trading. I will merge this PR anyways.

PR #1147, opened 2021-06-16
Add desktop launch icon during installation on macOS 

orenku:
@harrellbm  - My first PR in Superalgos
I've had to use an external npm package for setting the icon. I install it just for setting the shortcut icon and then it is uninstalled as part of the script. 
lmk whet you think.
Tested on macOS Big Sur v11.4


harrellbm (approved):
This looks great! Now you can start looking into adding a launcher shortcut and then the shortcut feature will be complete!

Luis-Fernando-Molina:
Can I merge this @harrellbm ?

harrellbm:
Yes this is good to go 👍🏼

PR #1149, opened 2021-06-17
Sync trade calculations if the data is present

ssplatt:
Coinbase Pro removes the order from its records after it is closed so there is no way for Superalgos to get information about it, which returns an empty `order` object. From there we must assume none of the values within Superalgos have changed so no new order synchronization is necessary.

### DIFF
--- a/Projects/Foundations/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/APIs/ExchangeAPI.js
+++ b/Projects/Foundations/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/APIs/ExchangeAPI.js
@@ -98,6 +98,13 @@
             tradingSystem.addError([tradingSystemOrder.id, message, docs])
 
             logError("getOrder -> Error = " + err.message);
+            if ( /[0-9a-z]+ NotFound/.test(err.message) ) {
+                logInfo("getOrder -> NotFound, so order is actually closed.")
+                order = {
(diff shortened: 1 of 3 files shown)


Smidy13:
@ssplatt has this been tested against any other exchanges? 

I don't have my Trading System set up after Beta 10 and the re-factoring therefore I can't live test this at the moment. I can start to work on this, but just curious if you've tested against Binance and Coinbase Pro or just Coinbase Pro? If not let me know and I'll focus on getting back to a Live Trading state to do some testing.

The code seems solid, its just adding in some IF statements on the OrderCalculations.js which will continue to return fine for Exchanges which return the order but in a cancelled state.

Also in the TradingOrders.js does Coinbase Pro return a order.status === 'NotFound', in that case are we still safe to handle the 'null' option too and close the order for this case?

ssplatt:
I have not tested against any other exchanges. The only other account I have which I can test in BinanceUS. I can try to make some other accounts if there are any specific exchanges you have in mind.

Null wasn't the real return. That was added by @Luis-Fernando-Molina as a first attempt but it was actually "undefined" until I faked the turn by manually creating the order object with status NotFound and all other expected data missing, or undefined.

Luis-Fernando-Molina:
Once you guys fell Ok with the change let me know to merge it. Since we are starting a new version, there is time for people using the develop branch to see if this is working. You just need to pay attention to complains of something not working to see if it is not derived from this change.

PR #1163, opened 2021-06-18
Improvement: Icons and Config for OKEx, Bug Fix: Status Text Offset

Smidy13:


Smidy13:
@Luis-Fernando-Molina the new statusAtAngle hasn't been set for all node types. Therefore I figure it is best to set this to 'false' by default and it can be changed to 'true' for new node types as required.

I've tested and with this change the Governance and User Profiles seem to still be good (as they change it to 'true' as required).

Luis-Fernando-Molina:
ok, I left them true thinking that at the Trading Engine they are displayed at angles, but probably they are farced to do so there too. So all good!

PR #1197, opened 2021-06-22
fix scroll bug on Mac

orenku:
Currently, there is a bug when using the mouse scroll button on MAC where the mouse is a standard one and doesn't support 'continuous scroll'
In continuous scroll, a scroll on the touchpad will result in a burst of mouse weel events. This is the reason the existing implementation uses some filtering mechanism (defined by MAC_SCROLL_SENSITIVITY & MAC_ZOOM_SENSITIVITY)

However, when connecting a standard mouse each mouse wheel tick results in a single mouse event, and the current MAC filter causes the scroll not to be responsive - meaning the user must scroll a lot in order to zoom in a single step. 

The suggested fix is based on the event behavior described in https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event

"._..On Mac, the value is complicated. The value is changed if the device that causes the native wheel event supports continuous scroll.

If the device supports continuous scroll (e.g., trackpad of MacBook or mouse wheel which can be turned smoothly), the value is computed from accelerated scroll amount. In this case, the value is the same as Safari.

If the device does not support continuous scroll (typically, old mouse wheel which cannot be turned smoothly), the value is computed from non-accelerated scroll amount (120 per notch). In this case, the value is different from Safari.

This difference makes a serious issue for web application developers. That is, web developers cannot know if mousewheel event is caused by which device._"

The suggested solution is based on the fact that non-accelerated scroll devices have a scroll amount which is a multiplication of 120, and it hardly possible to get to a value of 120 when using a 'continuous scroll' device (I couldn't get to a value above 10  
In this fix on Mac only devices, we check if the value of the wheel delta < 120, in that case, we assume 'continuous scroll' otherwise we skip the threshold mechanisim


Luis-Fernando-Molina:
You really made my day with this fix! Thank you so much, you can not believe how old this bug is in here and noone including me figured out yet. I will stop right now what I am doing and get out to celebrate this. ;-)

PR #1247, opened 2021-06-28
Contribution: Add EMA20 of the on balance volume

Raplh2502:


Raplh2502:
I need to clean up a little bit before I re-PR. I Cannot delete 27 remaining files.

PR #1258, opened 2021-07-03
Added sandbox mode

benmarte:
Please review and sorry for the formatting changes I'm not sure what you are using for formatting but I'm using prettier, let me know if that is an issue.

Thanks.


Luis-Fernando-Molina:
This is a very nice contribution, thanks!

JSON files are saved with 4 indentation spaces from within the system. I guess the formatter is configured in the same way because it usually matches what the system saves.

PR #1292, opened 2021-07-13

Raplh2502:
Adding 2 new indicators based on On Balance Volume / an integral comparison between ema3 and ema20 of the OBV to create the "Integral Oscillator" and a comparison of the current integral of ema3 and ema20 OBV and the average value of the current trend to create the On-Balance Volume Span


Raplh2502:
Unwanted modification in the launch.js... closing the PR

PR #1309, opened 2021-07-18

end-debug:


Luis-Fernando-Molina:
Hey I have some bad news with this. Together with the review, many other files were sent. This is not your fault, it is just that we are developing a feature to save plugins and we are in the process of updating the workspaces, and for some time all plugins are saved together with the workspace (until we finish this upgrade, then it will be optional). 

The thing is that since you sent your review from the develop branch you also included all data mines modified and I should not merge that. 

To fix this, copy the content of your review or your review file, switch to the plugins-docs branch both for current and contribution branch, and paste the review there and resubmit the review with a new PR from plugins-docs. That should send only one file. Sorry for the inconvenience. 😅

PR #1351, opened 2021-07-31
Reference Auto Fix Feature

harrellbm:


Luis-Fernando-Molina:
Great job man! I think you graduated with this one! congrats!

PR #1447, opened 2021-08-09
Contribution: Goverance Github Server Validation 6 Test 3

Luis-Fernando-Molina:


Luis-Fernando-Molina:
This Pull Request could not be automatically merged and was closed by the Superalgos Governance System because the User Profile Id already exists and belongs to another User Profile on record. 

User Profile Id = "1fd95e41-4b22-48d5-be3a-904c37afcb11" 

 User Profile with the same Id = "Julian-Molina"

PR #1668, opened 2021-09-05
User profile updates

harrellbm:


Luis-Fernando-Molina:
hey! thanks for the update, but it seems you updated the app while we were testing with julian some dependencies fixes. Now all is working and tested, so please reverse the changes to the package lock.

harrellbm:
My bad just reverted the add now it should only be my profile updates

Luis-Fernando-Molina:
This Pull Request was automatically merged by the Superalgos Governance System because it was detected that the Github User " harrellbm" who submitted it, modified its own User Profile Plugin File and nothning else but that file. All validations were successfull.

PR #1758, opened 2021-09-14
Contribution: fix on hartbeat csv files and filehandler for CSV file handling

Ferrox85:


Ferrox85:
This is my first PR, and I saw I forgot some things and I am not very handy with GitHub yet :-)

You can drop the changes on all files except the following files (maybe I can do this to, but nog very handy with it..) (I saved a file with my editor and I have a plugin with corrects all spaces/tabs/etc, which causes a lot of code rectifications)

Projects/Foundations/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/Snapshots.js
changes on this file:
- adds only the first run a header to the to be saved content
- second run it fetches the CSV file
- adds the new content to it
- saves the file 

this way you have a snapshot of all trades for referencing

Projects/Foundations/TS/Task-Modules/FileStorage.js
changes on this file:
-skips JSON validation (because it is no json file) and returns the text

Luis-Fernando-Molina:
Great, but it is better that you resubmit only the things you believe should be part of the PR. My suggestion is that you reconfigure your editor to match the code style already at the project, otherwise each file you touch will come with everything changed. That includes not using the ; which is optional.

I will be probably easier if I close this PR and you resubmit a new one only with the changes, agree?

Ferrox85:
Yeah, this is probably a good idea!

I'll try to figure it out how the settings must be, and make a new PR

PR #1837, opened 2021-09-22
Fixet empty rows and undefined strategy values

Ferrox85:
Complete fix of the CSV snapshot generator. Implemented the trigger values and overwrite the exit values at exit state with keeping the trigger state values. Removed the empty lines in the file.


Luis-Fernando-Molina:
Thanks for the PR!

Ferrox85:
@Luis-Fernando-Molina ,

This works fine on my local machine where it is developed. but on the remote machine it crashes.. it looks like it can't write the file. Not sure where this occurs due to hard debugging this on the remote machine...

Ferrox85:
Never mind, was a wrong formula on my remote computer..

Issue #1870, opened 2021-09-25
Asset: docker/container deployment docs topic


harrellbm:
Explain the use of the docker image and in-depth documentation of container deployment

PR #2062, opened 2021-10-04
Update READMEs, add brewfile, add npm script aliases

ssplatt:
- add table of contents to main readme
- clean up a few sections of the main readme
- move docker directions and info out to it's own readme
- add a Brewfile to quickly install requirements on Mac OS with Homebrew
- add a couple quick scripts to `package.json`
  - the `package.json` can function similar to a `Makefile`
  - `npm run setup` will execute `node setup noShortcuts`
  - `npm start` will execute `node platform minMemo noBrowser` 
- change docker-compose to use `user` instead of trying to set uid/gid via env vars which didn't work to ensure file persistence works

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,3 @@
+brew "git"
+brew "node"
+brew "npm"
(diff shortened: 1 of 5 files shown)


harrellbm:
Resolves #1870

PR #2109, opened 2021-10-11
Contribution: (Translation of "Tutorial" docs-article until "Initial values")

gitforkgit:


harrellbm:
The red lines are what was previously in the file.  While the green lines are what is being newly added to the file.  This allows anyone to see what was there and what is being added at the same time.

PR #2115, opened 2021-10-12
Contribution: German Translations

mARTin-B78:


Luis-Fernando-Molina:
image

Hey! you are standing at develop and contributing into plugins docs. That is not good, since you are carring more that 1000 files changed from one branch to the other. You can continue working on develop, and contributing docs changes there. 

If you want you can be standing at plugins docs and contributing to plugins docs. But please don do it from one branch into the other.

PR #2155, opened 2021-10-17
Develop

piyanthak:


piyanthak:
json errors on those........as of today 5am utc. There were no PR's pending at the time. Time Doc files had bad jsn formats "ES"

piyanthak:
Guess PR#2152 addresses the issue. Leaves just the updated workspaces to submit.

harrellbm:
You can use GitHub Desktop to roll back those last four commits and then everything should be good to go

PR #2159, opened 2021-10-17
Small fixes for debugging

alexandrustefan:


alexandrustefan:
Yes. And I think that the reason why this concatenation is not good anymore is due to recent changes added to the Environment.js : 
`if (process.env.PACKAGED_PATH) {
    basePath = process.env.PACKAGED_PATH
} else {
    basePath = __dirname
}`

Luis-Fernando-Molina:
Great! Let me know if you need ideas on more things to improve / add  :-)

PR #2174, opened 2021-10-20
Contribution: Accept webSocketsExternalURL parameter on network node config

ph4z:
When hosting the platform on a SSL server, websockets requests must be SSL too. The purpose of this PR is to accept an additional "webSocketsExternalURL" parameter provided by end user for activating wss.


Luis-Fernando-Molina:
Great! Can you please explain this feature at the docs too?

PR #2359, opened 2021-10-30
Contribution: (Translation into german: Short part of governance)

gitforkgit:


gitforkgit:
It somehow wants do delete the spanish part i think? Why and how can i fix it?

PR #2498, opened 2021-11-03

Raplh2502:


Raplh2502:
Submission of a brand new datamine with Mean breakout Indicator. This will be the 1st indicator of a strategy dedicated datamine

PR #2612, opened 2021-11-12
Contribution: Translation into Russian Getting Started Tutorial (step 60-65)

MarinitaMarina:


harrellbm (changes requested):
It looks like some icon changes got snuck in with your translations. Make sure that the icons are showing up right within the tutorial before we merge this

PR #2618, opened 2021-11-13
Contribution: (translation into german: "App contribute command")

gitforkgit:


harrellbm (commented):
Make sure this icon changes is correct otherwise this looks great!

Luis-Fernando-Molina:
@harrellbm I noticed those Icon changes at other PRs, it looks like they are involuntary and unrelated to the translating work being done. You should find the root cause of them, looks like a bug.

Luis-Fernando-Molina:
At a previous PR I manually fixed them, but now I am not sure if the system is not auto correcting an icon definition that is wrong, or is corrupting something that is right. Before more manual stuff someone should check that out.

Luis-Fernando-Molina:
Shall we merge this?

harrellbm (changes requested):
Are you submitting things from a new getting started workspace? I haven't been able to track down why icons are getting changed like this on accident still

gitforkgit:
You mean i should app.update since i wasnt working on superalgos for the last couple of days?
Shall i update and then resubmit?

harrellbm (approved):

harrellbm:
Perfect that cleared it up!

PR #2621, opened 2021-11-14
Improvement: Obtain Wallet and Liquidity Balances via WEB3

Smidy13:
@Luis-Fernando-Molina and @PLuv given this affects the token distribution wouldn't mind both of you having a quick look over this to ensure the code and logic is up to scratch.


Luis-Fernando-Molina:
@Smidy13 ok will check it out before the next distribution :-)

PR #2659, opened 2021-11-16
set brokerid for okex

ssplatt:
use a static broker id for okex, overriding the default value provided by the CCXT library. This value can also be overriden in the workspace by adding a `"brokerId"` field to the API Key json.

### DIFF
--- a/Projects/Algorithmic-Trading/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/APIs/ExchangeAPI.js
+++ b/Projects/Algorithmic-Trading/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/APIs/ExchangeAPI.js
@@ -24,6 +24,7 @@
         tradingSystem = TS.projects.foundations.globals.processVariables.VARIABLES_BY_PROCESS_INDEX_MAP.get(processIndex).SIMULATION_STATE.tradingSystem
   
         exchangeId = TS.projects.foundations.globals.taskConstants.TASK_NODE.parentNode.parentNode.parentNode.referenceParent.parentNode.parentNode.config.codeName
+        options = TS.projects.foundations.globals.taskConstants.TASK_NODE.parentNode.parentNode.parentNode.referenceParent.parentNode.parentNode.config.options
   
         let key
         let secret
@@ -37,7 +38,6 @@
                 secret = TS.projects.foundations.globals.taskConstants.TASK_NODE.keyReference.referenceParent.config.secret
                 uid = TS.projects.foundations.globals.taskConstants.TASK_NODE.keyReference.referenceParent.config.uid
                 password = TS.projects.foundations.globals.taskConstants.TASK_NODE.keyReference.referenceParent.config.password
-                options = TS.projects.foundations.globals.taskConstants.TASK_NODE.keyReference.referenceParent.config.options
             }
         }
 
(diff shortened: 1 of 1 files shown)


MatthewDin:
Hard coding exchanges in an 'API' is not wise. We've noticed param passing flaw and can be 'corrected' generically for future use.
Line 16:     let options = {}
Line 52:            options: options

And brokedId is okex specific so below is odd.
New Line 35: brokerId = TS.projects.foundations.globals.taskConstants.TASK_NODE.keyReference.referenceParent.config.brokerId

Solution: Assign non defined generic params (everything other than code, keys, sandbox) to 'options'. Code necessary to make this scalable and flexible and what CCXT expects to recieve.

Each CEX has their own unique list of parameters and this method ensures the Node config is able to handle ALL.

piyanthak:
I think you found a flaw. But unwise to hard code exchange in an API.

Use 'options' variable to pass any non generic exchange configurations ( let options = {}, options: options, all in ExchangeAPI.js)
This is long term solution since exchanges vary vastly in the params they require and optional.

ssplatt:
A flaw in what? Current method is that SA does not change the broker I'd so the underlying hard coded ccxt brokerId is used. So the new way is that SA uses its own hard coded brokerId but also provides a user a way to override the value with their own ID if they need to without modifying the code.

piyanthak:
Flaw that 'options' passed to CCXT is blank on ExchangeAPI (options = {}) . The 'idea' is that 'options' parameter array is used to pass all non generic exchange parameters. Thats the API 'standard.
But hey do as you see fit and patched to work for now. And wait for other exchanges to come online so this will be made more evident.

ssplatt:
This was definitely quick and dirty and gets the job done. So far nothing else seems to be using the options hash but I could see value in making more along the lines of a JSON.parse in the code as opposed to looking for specific values. However, I don't see a way around hard coding the SA brokerId as a way for all SA users to gain it's benefit and overriding the similarly hard coded ccxt value.

piyanthak:
'options' hash is parsed at okex.js for a lot more than brokerId. ;-)

ssplatt:
I meant that it doesn't appear that any other exchanges within SA are passing options into the hash as far as I can see from the code. Not that ccxt doesn't have plenty of options available which SA could be setting.

ssplatt:
pushed a change, should make it more generic but is untested

Luis-Fernando-Molina:
My suggestion @ssplatt : 

image

Let the brokerId by yet another parameter that lives at the Exchange Node config. That way each exchange config can have the SA brokerId for each one (as we make more partnerships) and that is not part of the codebase. The change is just about reading the excchange config property like in the screenshot above.

ssplatt:
I'll look it over again but I didn't see where the exchange config was read into this file. Then again, the exchange codename must be read.

Luis-Fernando-Molina:
Check out my screenshot, the highlighted is the way to extract the properties from the exchange node at that point of the code

ssplatt:
Those highlights are from the API key node, which is where the 'options' also highlighted now lives.

Luis-Fernando-Molina:
ohhh, ok

ssplatt:
I moved the options hash/dict over to the Exchange node. so a possible config would now look like:

```
{
    "codeName": "okex",
    "options": {
        "brokerId": "mycustombrokerid"
    }
}
```

if you want to override the defaults.

PR #2676, opened 2021-11-19
Debian Setup Fix & WSL2 / VSCode Setup Info

diveyez:
Added more **README.md** documentation for setting up with Debian, Raspberry Pi's and WSL.

Added informational links with information on how to setup **Visual Studio Code** to connect to **Docker** through **WSL2** for allowing users to directly interact with **SuperAlgos** through **VSCode** with ease and freedom from requiring anything more than an IDE to operate and start contributing to **SuperAlgos**.

**Edited File:**
[README.md](https://github.com/Superalgos/Superalgos/files/7561753/README.md)

_**Discovery,
Using VSCode with WSL/2 leads to new abilities regarding the stability while live deploying and editing of code on Windows.
WSL2+ now allows containers to become VM's and vice versa.
I will be looking into providing a CI pipeline that builds a working WSL VM which automatically installs a remotely accessible interface as SuperAlgos ages.**_


README.md
```diff
@@ -120,20 +122,113 @@ You can use Safari or Google Chrome as your default browser. If you run into a b
...
-You may verify the installed versions with these commands:
+You may verify the installed versions with this command string:
 
 ```sh
-node -v
-npm -v
-git --version
+node \
+-v && npm \
+-v && git --version
```
ssplatt, line 135: the end result is exactly the same with this "one liner" which is broken up onto different lines.

README.md
```diff
@@ -120,20 +122,113 @@ You can use Safari or Google Chrome as your default browser. If you run into a b
 [Follow the Node.js package manager install instructions](https://nodejs.org/en/download/package-manager/) for your distribution to ensure you are getting the latest version of Node.js. Many distributions only maintain an older version in their default repositories. Python 3 is only required for running machine learning (TensorFlow).
 
 ```sh
-curl -sL https://deb.nodesource.com/setup_16.x | sudo -E bash -
-sudo apt-get install -y nodejs npm git python3
+curl -sL https://deb.nodesource.com/setup_17.x | sudo -E bash - && sudo apt-get \
```
ssplatt, line 125: users have reported issues with node 17 so it's best to stick with 16 right now unless you are willing to work through the bugs if you run into them

README.md
```diff
@@ -120,20 +122,113 @@ You can use Safari or Google Chrome as your default browser. If you run into a b
...
+install -y \
```
ssplatt, line 126: no real reason to attempt to turn this into a one-liner. especially if the command is broken up onto multiple lines, the copy-and paste is the exact same result.

README.md
```diff
@@ -120,20 +122,113 @@ You can use Safari or Google Chrome as your default browser. If you run into a b
...
+(NVM & NPM Fix)
+
+For this to work you will need to [use NVM to install and control node] (https://github.com/nvm-sh/nvm)
+
+- You will need to remove any versions of node already installed on Debian due to the repositories currently being out of date.
+- - __This is necessary before proceeding.__ 
+
+```sh
+sudo apt \
+remove nodejs -y && \
```
ssplatt, line 154: this needs to be cleaned up. the `&& \` is unnecessary as there is no other command after the `&&`. Also, why are all of the broken up into different lines at seemingly random places. If you really want these to be "one-liners", which have the same end result when copied and pasting multiple lines as they would without the &&, then break the lines so the `&&` starts the new line, or aim for another standard like 80 characters.

README.md
```diff
@@ -120,20 +122,113 @@ You can use Safari or Google Chrome as your default browser. If you run into a b
...
+
+```sh
+sudo apt \
+autoremove -y && \
+sudo apt autoclean -y
+```
+
+```sh
+sudo curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+```
```
ssplatt, line 172: is there any reason why all of these consecutive code blocks are separated into different code blocks?

README.md
```diff
@@ -120,20 +122,113 @@ You can use Safari or Google Chrome as your default browser. If you run into a b
...
+
+```sh
+sudo apt update \
+sudo apt upgrade && \
+apt autoremove -y
+```
+
+```sh
+cd
+```
```
ssplatt, line 199: did you mean for this to be `cd`? wrapped in a single back-tick so it read with the other text?

README.md
ssplatt: also, in terms of clarity and readability I do prefer these all to not be jammed into one-liners

diveyez: Debian specifically, has an issue with multiline copy paste. Therefore I use oneliners to stream line the process.

README.md
diveyez: To make it easier to read for new linux users. Nothing more.

README.md
diveyez: This was overlooked. Sorry about this.

PR #2706, opened 2021-11-23
Contribution: "Abided by the instructions of L. Molina"

Tyhote:


Luis-Fernando-Molina:
You need to check that you are not sending any other file, like in this case.
We know this was unintentional, and some people here are looking into avoiding this, but in the meantime, you need to remove that change from the PR.

image

markinho91:
Estimado Luis Molina:Estoy obteniendo el mismo error de que mi package json está queriendo ser enviado al servidor cada vez que intento realizar un app.contribute.Cómo puedo excluir este archivo?Muchas gracias!Atte.Marcos G. Rivas.El 23 nov. 2021 13:50, Luis Fernando Molina ***@***.***> escribió:
You need to check that you are not sending any other file, like in this case.
We know this was unintentional, and some people here are looking into avoiding this, but in the meantime, you need to remove that change from the PR.

Luis-Fernando-Molina:
@markinho91 the guys in charge of this are discussing how to avoid this right now at the support chat group:

https://t.me/superalgossupport

PR #2713, opened 2021-11-23
install package to global scope to stop changing the package-lock.json

ssplatt:
- users should `npm ci` to install the packages which are defined in the package.json and package-lock.json
- use of `npm install` updates the local package-lock.json which we do not want normal users doing unintentionally
- when installing packages that are not necessary for the actual application, they should be installed globally (`-g`)
- other minor whitespace changes


Luis-Fernando-Molina:
shouldn't you  modify the read me instructions if now no more node setup is needed?

PR #2756, opened 2021-11-27
Contribution: Add return and price probability analyzers to Polus data mine

Raplh2502:


Raplh2502:
I have also pushed an update of the depencies at Masters Trading Mine and minor polishing to Enhanced Bollinger

PR #2808, opened 2021-12-01
Contribution: "more spelling issues fixed"

Neobii:


Neobii (commented):
Ah dang, looks like it committed my x and y positions.  Does this need to be programmatically ignored?

PR #2869, opened 2021-12-03

end-debug:


.gitignore
```diff
@@ -21,3 +21,5 @@ package-lock.json
 .idea
 dist
 /nbproject/private/
+Data-Mining
+Algorithmic-Trading
```
ssplatt, line 25: these probably shouldn't be added globally.

PR #2884, opened 2021-12-04
Fix dragging nodes out of bounds

whtv:
The condition `(displaceVector !== undefined)` in `Canvas.js` was always true, so that a grabbed node would always be dragged when using `Ctrl + ArrowKey`, even after the floating space container had been stopped at the canvas bounds by the physics function. It was easy to accidentally (or on purpose :/) drag nodes out of reach and visibility that way.


whtv:
Also, `if (visible === false) { return }` is ineffective. Not sure what the idea was there initially so I left it in.

Luis-Fernando-Molina:
@whtv this is a very good fix!!! Finally someone did it!! 

Now that you are around fixing stuff, there is another annoying stuff happening, probably at some other issue:

We can not close a side panel when the tab-handle is on top of a node at the design space, because the node takes precedence on mouse clicks. Can you try fixing that one please?

whtv:
@julian-molina https://github.com/Superalgos/Superalgos/pull/2908

PR #2908, opened 2021-12-05
Give Side Space precedence on mouse click

whtv:


Luis-Fernando-Molina:
Great!

Luis-Fernando-Molina:
Have to reverse this change because it disable the workspace panel.

https://github.com/Superalgos/Superalgos/pull/2913

PR #2913, opened 2021-12-05
Contribution: Revert Bug Fix of side panels because it disables the workspace panel

Luis-Fernando-Molina:


whtv:
Not sure why that happened. Anyway, https://github.com/Superalgos/Superalgos/pull/2935 should be the proper way to do it.

PR #2991, opened 2021-12-07
Contribution: (updating claims)

gitforkgit:


gitforkgit:
@julian-molina Here seems something to go on, in the commit overview it shows the commits i did after my last pull request should have been merged, but when going to files there are only my translation files without my profile update.

PR #3047, opened 2021-12-09
Dependecyfilter And Remote Console Actions

Ferrox85:
Multiple Enhancements:

**Use of variable in for example the timeframe of an indicator**
It is possible with this addition to make formulas like this:
chart[sessionParameters.userDefinedParameters.config.timeFrame].candle.close
When the session parameter can be changed during the trading. So you can move timeframes with your strategy. For example if you use the same strategy, which also works on other (higher) timeframes, but do not want to recreate this whole strategy. you can set the variable when needed and return back to the start of you managed stage. The Dependecyfilter loads in all possibilities, so you are not missing any indicator if you want to change.

**Added Console actions.** 
It is now possible to interact (limited) with SA by console. The advantage of this is that you can write scripts (Shell or Applescripts) to interact on a certain level with SA.
These are the additions:

remoteRunSession: remoteRunSession
Call this, and you can initiate a backtest (or live trading) with the node id, example:
UI.projects.algorithmicTrading.functionLibraries.tradingSessionFunctions.remoteRunSession("node-id-of-session")

changeFormulaInNode: changeFormulaInNode
Call this, and you can change the formula in a node example:
UI.projects.algorithmicTrading.functionLibraries.tradingSessionFunctions.changeFormulaInNode("node-id" , "New formula")

changeConfigInNode: changeConfigInNode,
Works the same as changeFormulaInNode but then for a config of a node. For example the timeframe needs this:
"UI.projects.algorithmicTrading.functionLibraries.tradingSessionFunctions.changeConfigInNode("node-id", "config"
Where config can be: "{"label":"01-min"}

RunremoteParentSession: RunremoteParentSession
Runs a taks:
UI.projects.algorithmicTrading.functionLibraries.tradingSessionFunctions.RunremoteParentSession("node-id-of-taks")

**Fix in Snapshots**
Apparently Snapshots couldn't handle double (( or )). Fixed with ReplaceAll
There is stil a bug in combination with the ability of changing the timeframe dynamically, but is doesn't break SA


Luis-Fernando-Molina:
>  Multiple Enhancements:

> Use of variable in for example the timeframe of an indicator
> It is possible with this addition to make formulas like this:
> chart[sessionParameters.userDefinedParameters.config.timeFrame].candle.close
> When the session parameter can be changed during the trading. So you can move timeframes with your strategy. For example if you use the same strategy, which also works on other (higher) timeframes, but do not want to recreate this whole strategy. you can set the variable when needed and return back to the start of you managed stage. The Dependecyfilter loads in all possibilities, so you are not missing any indicator if you want to change.

This is great, please tell me because I could not easily guess it by reading the code:

1. The change will only be used if there is a sessionParameters.userDefinedParameters.config.timeFrame defined, right? I was not sure if there was any other session parameter used, the code would still trigger the replacements.

2. If users don't use sessionParameters.userDefinedParameters.config.timeFrame then nothing changes, right?

Luis-Fernando-Molina:
> Added Console actions.
> It is now possible to interact (limited) with SA by console. The advantage of this is that you can write scripts (Shell or Applescripts) to interact on a certain level with SA

Please clarify: when you interact from console, you do it with the Platform Client already running and the UI running? Or the UI does not need to be running at all?

The only way I can imagine this working is from the browser's console. Am I right?

Ferrox85:
> great, please tell me because I could not easily guess it by reading the code:
> 
> 1. The change will only be used if there is a sessionParameters.userDefinedParameters.config.timeFrame defined, right? I was not sure if there was any other session parameter used, the code would still trigger the replacements.
> 2. If users don't use sessionParameters.userDefinedParameters.config.timeFrame then nothing changes, right?

The code works like this
1:it looks for the use this string: "[sessionParameters.userDefinedParameters.config" (the [ implies that it is used for an object in the JS code)
2: It checks if it is used in this form: chart[sessionParameters.userDefinedParameters.config, and the content of the variable starts with "at" (as in at05min for example) to verify that you are using a variable timeframe.
3: if it does, it adds all possible timeframes to the array of code, it adds all possibilities to the list so they are in the filter. (for ex: chart.at01min.candle.close, chart.at02min.candle.close, e.t.c. with a loop)
4: if it doesn't, it simple replaces the UserVariable with the content of the variable. (I don't think you want to use it in this way, but it works. But if you change the variable during the trade it breaks the same way as it did before) So advice is to only use this with variable timeframes, or fixed at the start of the trade.

So, it does not need te be the exact name "timeFrame" (I used this for a test) but you can call it time, timeManagement or whatever. But if it is directly behind "chart" in the code node, and the content of the variable starts with "at", then this code is running.
It think I need to fix a minor bug, because I presume now that chart is at the first of a line.. but this is not the case when a exchange or market is in front of this. But this is a minor fix (change a === to >=) I wil do this if this PR is closed.

Ferrox85:
> > Added Console actions.
> > It is now possible to interact (limited) with SA by console. The advantage of this is that you can write scripts (Shell or Applescripts) to interact on a certain level with SA
> 
> Please clarify: when you interact from console, you do it with the Platform Client already running and the UI running? Or the UI does not need to be running at all?
> 
> The only way I can imagine this working is from the browser's console. Am I right?

The UI needs to be running indeed, and it is in the browsers console, correct. I think this is the most easy way now.. because there is a lot code needed to interact with SA for this moment I think. Most convenient way  is an API which does exactly this. But there are some things handled in the UI what I can't comprehend. So this way works fine for me, en possibly others too.

Luis-Fernando-Molina:
RE: Timeframes, ok, no further questions there.

Luis-Fernando-Molina:
RE Command line functions:

Ok, Now I understand what you are trying to do. I have a few suggestions though:

1.  Create a new module for these functions, like an utility module under the Visual-Scripting project. 
2.  Maybe instead of running these commands at the browser console, which no user is used to that, consider creating in-app commands for this like the existing commands set: app.update, app.contribute. You can create a new set of commands with a different context, instead of app, you can use ui for instance , because the app commands are in fact for other purposes. By creating these commands in-app, you will be able to document them well, and also see a results page if needed too.

Luis-Fernando-Molina:
By moving these functions to another module you would also be resolving the current merging conflicts.

Maybe you could remove the functions from the PR and leave the filter changes, so that I can merge this in the meantime, and also gets easier to review.

Ferrox85:
> By moving these functions to another module you would also be resolving the current merging conflicts.
> 
> Maybe you could remove the functions from the PR and leave the filter changes, so that I can merge this in the meantime, and also gets easier to review.

Yeah, no problem. And good suggestions too, but I don't know if that would work for others. Because invoking this with the gui like app.update is not possible with external scripts I think? I will move them anyway in the future to a separate module. (but this also produces the need of changing my own scripts... so that's a but of work.)

PR #3064, opened 2021-12-10
[Distribution] Link mobile wallet for user profile creation

NAmorim:
All eyes on deck to validate this... I've tested in iOS and Android using Trust Wallet and Metamaks.

CAUTION: There's a bug on Trust Walet in Android where you MUST select the wallet. If not you get an error about "Wallet not supported"


Luis-Fernando-Molina:
Don't forget to update the tutorials / docs, or get someone to do it, otherwise we will be flooded with questions at the support group or people just don't know about this way of creating the profile.

Luis-Fernando-Molina:
Also, maybe it would be good to split the current Build Profile menu item in 2:

Build Profile And New Wallet
Build Profile With Mnemonic
Build Profile With Wallet Connect

We could route the first 2 via the same current function Build Profile with little effort, but it would be clearer for the user point of view.

PR #3067, opened 2021-12-10
update slackbot to use nodejs module and function as a slack application

ssplatt:
- converts the slack bot from a simple webhook to a proper app, opens the door for slash commands
- converts send test message to use the actual code that the trading system will use while sending announcements
- also add doc pages for Slack and Twitter bot


ssplatt:
all of the deepscan issues are not from this PR

Luis-Fernando-Molina:
Beautiful!

PR #3072, opened 2021-12-11

markinho91:


markinho91:
> 

This is my new profile, with the new addrees linked. Only need to finish configurating the claim and vote node but all my referral, mentor and support programs are there

PR #3078, opened 2021-12-11
Update README.md

francescocambi:
Fix docker run paths, otherwise you'll have a permission denied error when the process tries to write on that paths.


ssplatt (approved):
👍

ssplatt (changes requested):
actually, just noticed this is a PR to `master`. I believe this is already fixed on `develop`

ssplatt:
i double checked, this is not fixed on develop, please redo this PR against the `develop` branch

PR #3083, opened 2021-12-11
update directory paths for persistent data in readme

ssplatt:


vraptor75011:
@ssplatt I think you forgot to update the lines creating local directories ?
mkdir -p Data-Storage Log-Files My-Workspaces

ssplatt:
you are correct. updated.

PR #3137, opened 2021-12-15
Contribution: (Translation into german: "LAN Network")

gitforkgit:


harrellbm (changes requested):
Hey it looks like your German is overwriting the Russian translations in the lan-network.json 

Do you remember how you edited that file? There may be a bug we need to find

gitforkgit:
@harrellbm I think i didn`t change the file itself at all in any text editor, i just translated it inside Superalgos. Didn`t me and others have this problem too before? I will look for where the problems could come from  from a users perspective and change  the error file.

harrellbm:
@gitforkgit I just went through the lan network file and there were so duplicates so I am guessing that the duplicates were probably the issue this time. But we should keep am eye out if it happens again to see if it's a sign there is something else wrong with the docs in that section.

Now you should be able to update your develop branch and replace the lan-network.json with the new on in the develop branch and then we'll get this set merged 👍🏼

gitforkgit:
@harrellbm I manually fixed the file, hopefully correct. If its ok from the structure and it gets merged, please afterwards ask a russian translator if all translation fit the position where i copy pasted them. :)

harrellbm:
@gitforkgit Nice now you've got the deep docs knowledge ;) you just missed one overwrite on line 12 otherwise everything else looks perfect

gitforkgit:
@harrellbm Alrighti, i will have a look after it.

harrellbm:
image

This is the line showing up for me as I review the request.  If it's fixed in your local instance you might just need to git push from your local instance

gitforkgit:
@harrellbm  Ohh, i think i found out why it seems to be correct in my editor but wrong here, i deleted it from the lines directly over the german translation but added it a bit upper. If you take a look at the file where its marked missing, have a look a couple lines further up, the russian translation is there, so its not missing but in a new order. If the order makes a problem let me know.

harrellbm (approved):
@gitforkgit Ah yes that is what is going on. The order doesn't matter so this is good to go 👍🏼

PR #3140, opened 2021-12-15
Better mouse click handling

whtv:
This assures that the `click` event is passed to the container that last listened to the `mousedown` event.


Luis-Fernando-Molina:
Awasome!

Issue #3163, opened 2021-12-18
gnome-terminal hard coded but not standard


isysd-mirror:
## Context

*Complete and adjust the contents leaving the information that fits your case and deleting the information that doesn't.*

**Operating System:** Linux (Xubuntu)

**Software version:** today's master branch

**Workspace version:** default Workspace.json (Getting Started Guide)

**Reproducible:** the issue may be reproduced.

### 1. Explain what the problem seems to be.

Installing Superalgos using `node setup.js` installs a desktop shortcut which assumes that gnome-terminal is also installed. This is not the case for me, nor for any non-gnome user.

### 2. Clearly explain what you wish to accomplish.

A more generic terminal launch command should be used. There's no reason why this can't work for all linux users with a line of code changed. Specifically this line:

https://github.com/Superalgos/Superalgos/blob/718cc983e53d26156600f184022fd8c37738a669/Launch-Scripts/create-shortcut.js#L62

PR #3164, opened 2021-12-18
patch for desktop launcher bug GH #3163

isysd-mirror:
This change fixes bug #3163, regarding hard coding `gnome-terminal` into the desktop launch command. I believe it is unnecessary to specify a terminal, and simply removed the explicit reference. Confirmed that this patch fixes the issue for my local Xubuntu install.


Luis-Fernando-Molina:
Welcome back!

PR #3170, opened 2021-12-19
Contribution: (Translation into german:"WebGL Backend")

gitforkgit:


harrellbm (approved):

harrellbm:
@gitforkgit It looks like someone has been working on the same files as you and there are some conflicts

gitforkgit:
@harrellbm So i have resolved the conflicts, i think. It is a bit weird that there is more trouble the last days, often with russian translations. Is it just because two languages commited changes or a Bug?

PR #3174, opened 2021-12-20
Added the ascendex exchange logo to UI

cozed-gh:
image


cozed-gh:
Logo add for ascendex.

Bugfix:
On the event of backtesting I found that I was thrown an error in the logs of having reached the minimum Base asset balance, because my minimumBalance was set to 0.
The comparison for the of the values is:
                        if (tradingEngine.tradingCurrent.tradingEpisode.episodeBaseAsset.balance.value <= sessionParameters.sessionBaseAsset.config.minimumBalance) {

but changes only that to be a < instead of <= to allow for the case of the strategy of buying a Base Asset, starting off from 0, and having minimum balance of 0.

This might be wrong when the case is like this - But I figured that the 'Getting Started' is currently set up like this and throws this error.

Luis-Fernando-Molina:
@cozed-gh @julian-molina I haven't seen yet an Exchange Logo in which we include the exchange name. Wouldn't it be better to remove the name form the icon for consistency?

PR #3178, opened 2021-12-20
Update package-lock.json

end-debug:


ssplatt:
why is the package-lock.json being updated? is there something specific you are trying to accomplish?

Luis-Fernando-Molina:
@end-debug please remove the package json from your commit. @ssplatt shouldn't package json be git ignored?

ssplatt:
it is in the git ignore but i'm not entirely sure why that is not respected. It shouldn't be ignored, really, and we should fix whatever automated command (app.update?) that keeps updating it and including it in the PRs.

end-debug:
This usually happens when I do a npm update and some dependencies get
updated plus I don't know how to remove commits.

still Github desktop and stashing somethings before committing

Luis-Fernando-Molina:
get the latest package-lock.json from the repository develop branch into your local repository and do an app.contribute again. That will fix it.

PR #3196, opened 2021-12-21
Download external scripts at setup

whtv:


setup.js
```diff
@@ -137,3 +143,23 @@ function tfjsWinInstall() {
     console.log("Doing this will cause a pop-up script in a new window which will install many npm/node tools and dependencies for use within the WinOS ecosystem.");
     console.log("2.) When the installations finally complete you need to navigate back to your Superalgos directory and run `node setup` again.\n");
 }
+
+// Donload external scripts
+console.log("");
+console.log("Downloading external scripts …");
+console.log("");
```
ricardopoleobackup, line 150: Since in previous lines of this same method, we use `\n` to add additional vertical lines, what do you think about changing these console.logs into something like:
`console.log("\nDownloading external scripts …\n");`
So we are consistent within the methods.

ricardopoleobackup: Oh well, fair enough, I reviewed most recent changes in this file, and it looks like a coding style thing (single quotes for empty console and double for text with content).

I +1 the file rework of the whole file, tho.

Luis-Fernando-Molina: happy to see someone spontaneously reviewing PRs.

PR #3197, opened 2021-12-21
Changed to be only the Icon of AscendEX

cozed-gh:
image


ricardopoleobackup:
Any particular reason to change the condition in the method ´checkMinimunAndMaximunBalance´ from ´<=´ to ´<´ (in the ´TradingSimulation.js´ file)?

It doesn't looks like it is related to the scope of the PR (that targets the Icon)

Luis-Fernando-Molina:
@cozed-gh you committed a code file together with the icon. Was that intentional?

Projects/Algorithmic-Trading/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/TradingSimulation.js

cozed-gh:
> Any particular reason to change the condition in the method ´checkMinimunAndMaximunBalance´ from `<=` to `<` (in the `TradingSimulation.js` file)?
> 
> It doesn't looks like it is related to the scope of the PR (that targets the Icon)

I know I forgot to mention it in the topic. Read the reply to Luis to get up to speed on why I'd like that to be changed.

cozed-gh:
> @cozed-gh you committed a code file together with the icon. Was that intentional?
> 
> Projects/Algorithmic-Trading/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/TradingSimulation.js

It is intentional.

Reason for that change is that it will reduce the risk of having to deal with "silent" error messages (when the minimumBalance has been reached, it will only be logged in log-files and shown as a warning as a chart, never mentioned at the console).
Also, from my point of view, I found it proccing while having it set to '0', while running your Trading System having any condition that can evaluate to 'false' to trigger your Take Position.
The set back is basically that it will proc when the asset balance reaches below the minimumBalance - If it is set to 0, it has to reach some sort of negative value. If it is set to 1 BTC, it has to reach below 1 BTC. Given that this function is run when there is no strategy chosen - so basically inbetween 'Close Stage' and 'Trigger Stage' - which means that a trade would have to take place before being read. In the case of having set the minimumBalance is set to 1 BTC, the chances of your balance reaching 1.000000000000 BTC is quite small and wouldn't be a big change to have it proc at 0.999999999999 BTC or lower instead, with the reduction of errors.
Although I don't see any use case for keeping the minimumBalance at the session if you are accepting '0', I've seen and spent hours of time to help others that havn't learned that it is optional and have it there by mistake, because it was included in their base workspace that they started to build upon.

PR #3207, opened 2021-12-22
Contribution: Translators Team-Greek-PART of: âChapter II - Design Spaceâ,"Chapter IV - Charting Space"

theobserver:


harrellbm (changes requested):
Hey it looks like you are overwriting some of the other translations. Try updating your fork and see if that takes care of the overwrites

prepheadrus:
> Hey it looks like you are overwriting some of the other translations. Try updating your fork and see if that takes care of the overwrites

Hi Brenden, I don't know if you noticed. But this has started to emerge a lot lately. I had the same problem, or tiger40 also has the same problem. Could the error be elsewhere? I don't think it makes sense that we all make the same mistake.

theobserver:
Having been the example (twice) of how a translator can mess it up with pull requests , app.updates , problems with merging due to various errors e.t.c. is quite uncomfortable.
Not to mention the time spent trying to understand github and fix my errors using it instead of translating.

So i will start from scratch with the knowledge i acquired from the various helping posts and YouTube videos the kind mates here write/make. Hopefully i will make it to the end without any other errors (which i really doubt) since i reckon it will take more time trying to fix and more burden from mates trying to explain.

Sorry for any inconvenience .

Regards,
Nik.

P.S. Maybe it would be more productive if translators had only a file (e.g. word or any other kind) to translate and there was a coordinator (or similar) with sufficient knowledge of github that would complete the transition of the translated language files in superalgos, instead of trying to explain/educate all noobs like me.

PR #3210, opened 2021-12-22
Make tutorial draggable + fix top space click-through [needs setup.js to be run!]

whtv:
Use jQuery UI on the tutorial DIV to make it draggable.

_This requires a setup.js run (`node setup`) to get the external scripts, cf. https://github.com/Superalgos/Superalgos/pull/3196._


isysd-mirror (commented):
This commit is broken. It creates no requirements for jquery, but does not actually include the files which are required.

```
jquery-3.6.0.js:1 Failed to load resource: the server responded with a status of 404 (Not Found)
jquery-ui.js:1 Failed to load resource: the server responded with a status of 404 (Not Found)
```

PR #3213, opened 2021-12-22
Contribution: testing team member delegation

gustavohcorradi:


gustavohcorradi:
i do not understand the conflict, but it was only a test, so i can close the PR.

PR #3272, opened 2021-12-28
Contribution: (Translation into german: "Screen Capture and Sharing")

gitforkgit:


BastianMuc (commented):
Hello @gitforkgit , at first sight, your commit appears to overwrite existing RU & ES translations. Is this intended?

gitforkgit:
@BastianMuc No definetely not. This sometimes happens but @harrellbm and I couödnt figure out how this can happen. I didnt change anything since the last commits that got merged without a problem. I also app.update regulary...Tomorrow i will have a look and fix the files but i dont know where the problem comes from.

harrellbm:
@gitforkgit I think I have figured out the problem.  Even doing app.update regularly, may not be "often" enough.  If a different PR adds a translation to the same file and gets merged before your PR (even by a few seconds) then this kind of conflict will pop up.  Since we are all editing the same docs files there is no real way to total avoid it other than making sure to update your fork whenever you see this pop up

gitforkgit:
> @gitforkgit I think I have figured out the problem. Even doing app.update regularly, may not be "often" enough. If a different PR adds a translation to the same file and gets merged before your PR (even by a few seconds) then this kind of conflict will pop up. Since we are all editing the same docs files there is no real way to total avoid it other than making sure to update your fork whenever you see this pop up

You mean going to my own fork and click the "fetch and merge" Button?

gitforkgit:
When I go to vs code usually it shows me the conflicting parts but after I app.update and opened vs code there is no "error". How can i fix the problem or is the problem already fixed? Please let me know so i can keep translating after this "problem" is solved.
Thank you and happy new year if you read it on new years day. :)

harrellbm:
@gitforkgit thanks and same to you! 

this is a weird one.  I see that you updated your fork which should have done it but it looks like instead of git picking it up as a conflict it is taking it as a "replacement".  The test you did is coming through fine it's just those four other files. So you will probably have to replace them with the old ones and then copy and paste in your translations again.

maxah:
> When I go to vs code usually it shows me the conflicting parts but after I app.update and opened vs code there is no "error". How can i fix the problem or is the problem already fixed? Please let me know so i can keep translating after this "problem" is solved. Thank you and happy new year if you read it on new years day. :) @harrellbm

If conflicts arise, I usually re-do 'app.update' then 'app.contribute'. If there are conflicts again, but if there aren't too many conflicts, I fix them right on the GitHub site and ask for a merge there as well. You can also temporarily move the conflicting changes to another branch (I do this in 'GitHub Desktop'), then do 'Pull request', usually after this there are no conflicts. Then you pull back what you pushed to the other branch and do the 'Pull request' again. That's it. Happy New Year!

gitforkgit:
As my usual fixes dont work here i will follow your advice harrel as it is the easiest for me. I will manually copy the files from github replace them and insert my translation altough it is a bit annoying. 
@maxah Thanks for your advice. As I am on Linux and github desktop isn`t available there originally i cant do it this way.
Nevertheless thanks both of you @maxah @harrellbm . @harrellbm could you close this PR so i can keep going translating and later on insert this translation that is conflicting.

PR #3326, opened 2022-01-02
Contribution: (Translation into german and small changes: "Telegram Groups")

gitforkgit:


gitforkgit:
@harrellbm seems like the same problem

harrellbm:
@gitforkgit It looks like your fork is just enough behind for the issue to pop up

Capture

either that or we need to figure out a better way for the system to not think your new translations are replacing the others

gitforkgit:
@harrellbm it should  work now after copying file from the girthub repo....because of the errors i couldn`t update regulary and then files where deleted in the develop branch, but now all should be fixed up

PR #3515, opened 2022-01-05
Created Open-Storage Icons & SVGs

end-debug:


teehanming (changes requested):
Very nice! I only need you to change a bit the "available storage" icon. The tick need to be more formal-looking.

PR #3580, opened 2022-01-16
Contribution: "added colors to masters supports and resistances panel"

Neobii:


Neobii:
@Raplh2502 Did this come out right?

PR #3596, opened 2022-01-18
Develop-Icons-Update 

DJ05GaZ:


gitforkgit:
@DJ05GaZ I looked trough the new Icons, i like them all. I just saw at least I think so a small mistake at the last File: Projects/Workspaces/Icons/workspaces.png    On the right side there are corners left. If Iam wrong or that is intented to be like that, I am sorry.

PR #3601, opened 2022-01-18
Contribution: User Profile maxah

maxah:


Luis-Fernando-Molina:
This Pull Request could not be automatically merged and was closed by the Superalgos Governance System because the User Profile is not in a valid JSON format.

 err = "Unexpected token @ in JSON at position 0"

PR #3670, opened 2022-01-24
First update

jobidon:
hopefully it'll work better this time...


harrellbm:
@jobidon Hey what is the motivation for taking out all the exclamation marks?

You need to be careful changing the titles of articles because that is how the system finds them to display in tutorial, or inclusion in books, etc. 

This looks like a lot of good work! Just want to make sure we don't make a mess if unwanted changes make it through.

jobidon:
... not ALL, but many! :)
It improves readability and allow emphasis to be put on certain points: When used everywhere, the emphasis is lost.

jobidon:
I just realized that perhaps changing the label of the Expand button will break many things... This needs to be investigated further before being pulled to the main repo

gitforkgit:
@jobidon do you want this pull request closed until you know if and how it will break things or open?

PR #3673, opened 2022-01-24
Contribution: (small typo fixed: "General Notions on Documenting")

gitforkgit:


gitforkgit:
@harrellbm that was a pragraph without a real text in it,I think just in case the author wanted to add a paragraph.

PR #3693, opened 2022-01-26
Update to docs, tuts and UI

jobidon:
Removed many exclamation points from the tutorials
Fixed shortcut generation in linux
and updated the label on the expand button


harrellbm:
@jobidon Hey for the sake of code review can you split this PR into smaller pieces so that changes can be evaluated by the proper code reviewers?

gitforkgit:
@jobidon is that possible? As it would help delegating the single pull requests to the extra teams.

jobidon:
sure. Should I close this PR and submit each commit as its own PR?

gitforkgit:
@jobidon It would be great if you could make different pull requests for each team. So put all translations in one pull request. Put each "Fixed shortcut generation in linux" into one and "updated the label on the expand button" into one. Remember app.update before you make the new pull requests. That would be great. :+1:

PR #3694, opened 2022-01-26
German translation Foundation > Nodes > Open Execution / Topics > How…

lindersi:
… to attach a node... and Introduction 001-006


lindersi (commented):
Everything correct. Deletions only in one case of a translation update.

gitforkgit (approved):
All files seemed to be good :+1:

PR #3726, opened 2022-01-30
Contribution: Education - Translation - Ru - Data Mines - Most Indicators

maxah:


gitforkgit:
@maxah  Did you want to add these files? : Plugins/Data-Mining, Plugins/Portfolio-Management. And as a context as this Pull request is huge, did you add new articles?

maxah:
> @maxah Did you want to add these files? : Plugins/Data-Mining, Plugins/Portfolio-Management. And as a context as this Pull request is huge, did you add new articles?

Only translations. This new submodular system makes an endless mess.

PR #3778, opened 2022-02-05
Education: Renaming duplicate page titles - fixes #3554

### LINKED ISSUE #3554: 3. Tutorial Step - Client "Next" Button does not work
Hitting "Next" leads to the same page you are on.. so you can not go to the chapter "4. Tutorial Step - Collection of Servers"

BastianMuc:


harrellbm (changes requested):
@BastianMuc Hey thanks for updating this. We need to make this article more specific now that the ecosystem is developing.  This article should be titled Superalgos Platform Client rather than servers.  The platform client is the collection of server side code that runs the platform.  Which needs to be differentiated from the individual server components as well as the p2p network servers. That should help make things a little more clear for people

BastianMuc:
@harrellbm Page title is changed as requested.

harrellbm (approved):

PR #3794, opened 2022-02-07
Documented the Turtles mine

quantum8:


Projects/Foundations/Schemas/Docs-Books/community-data-mines.json
```diff
@@ -56,14 +56,18 @@
         {
             "style": "Chapter",
             "text": "X->Foundations->Topic->Sparta Data Mine->"
+        },
+		{
+            "style": "Chapter",
```
gitforkgit, line 61: @quantum8 If you have a moment, could you move the opening bracket  to the left so it lines up with the closing brakcet above? It's not important for the functionality but the structure.

Projects/Foundations/Schemas/Docs-Topics/T/Turtles/Turtles-Data-Mine/turtles-data-mine-001-turtles.json
```diff
@@ -0,0 +1,83 @@
...
+            "style": "Title",
+            "text": "Turtles on the Charts"
+        },
+		{
+            "style": "Text",
+            "text": "This trading system is a trend following system that attempts to catch a breakout. It is similar to a Donchian Breakout strategy."
+        },
+		{
+            "style": "Link",
+            "text": "Turtles System 1(investopedia)->investopedia.com/articles/trading/08/turtle-trading.asp"
```
gitforkgit, line 19: Here the same.

gitforkgit (approved):
@quantum8  @harrellbm  Everything looks quite good. There are only two small things that would look better. If you don`t want to or don`t have the time to change it, just let me know. After it getting merged, i can make the small changes as well. :+1:

Projects/Foundations/Schemas/Docs-Topics/T/Turtles/Turtles-Data-Mine/turtles-data-mine-001-turtles.json
quantum8: Just had a look and this is interesting @gitforkgit . I've been using templates from existing docs from way back and haven't noticed what's going on until now; I use notepad++ and it looks like my default setup is 4 spaces = 1 tab, whereas github here must parse it differently. The template I use has 2 tabs at opening bracket and 8 spaces for the closing one. 
What shall our standard here be?

gitforkgit: @quantum8 @harrellbm You once commented what the standard is or was at that point. Do you know what the standard is right now?

harrellbm: I believe it has been 4 across other places in the system. So let's go with that

PR #3802, opened 2022-02-09
Fix Bug: Multiple check-in attempts by Trading bot will only be respo…

PLuv:
…nded to once per candle by Portfolio-bot


PLuv:
Nevermind, will check from Trading Bot task and leave PM to send multiple check in confirmations.

PR #3813, opened 2022-02-10
Setup unit tests

Char1zardd:
Here is a summary of changes at a high level

- **package.json**: added a section for jest config, made it so that jest will work with ESM scripts should we decide to use them in the future. Also added a script to run tests, if making new tests I recommend using `npm run test:watch` so it hot reloads
- **__tests__**: this is where our unit tests will go. We can split this further into unit and e2e folders later, but I think for now we should focus on getting unit tests on all the major `.js` scripts first. You will notice a unit test for the `Launch-Scripts/system-check.js` file
- **Launch-Scripts/system-check.js**: I made some minor tweaks, such as pulling in the required node, git and npm versions from `Environment.js` to make tests more robust
- **Environment.js**: Added npm, git, and node minimum versions here


Char1zardd:
Deleting this PR to ensure I pull the right changes first, will open a new PR

PR #3836, opened 2022-02-16
Update dockerfile

vcaldas:
Updates base image for SuperAlgos container.

Issue:
using the previous version, the application hangs on the UI (always loading).

Tested on:
Rpi 4 4Gb running portainer (armv8)

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -1,8 +1,8 @@
-FROM node:16-alpine
+FROM node:17-alpine3.14
 WORKDIR /app
 COPY . .
 RUN apk add --no-cache --virtual .build-deps make g++ \
-    && apk add --no-cache python3 \
+    && apk add --no-cache git bash python3 \
     && node setup noShortcuts \
     && node tf_setup \
     && mkdir -p ./Platform/My-Data-Storage ./Platform/My-Log-Files ./Platform/My-Workspaces \


vcaldas:
Solved by #3835 
Duplicate.

Docker/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM node:16-alpine
+FROM node:17-alpine3.14
```
ssplatt, line 1: i think we want to stay on 16 as it is the LTS version. we've had other bugs pop up by running 17 before.

Docker/Dockerfile
```diff
@@ -1,8 +1,8 @@
...
 WORKDIR /app
 COPY . .
 RUN apk add --no-cache --virtual .build-deps make g++ \
-    && apk add --no-cache python3 \
+    && apk add --no-cache git bash python3 \
```
ssplatt, line 5: we don't want git in the container because the code should not be changing. these containers are aimed at production deploys.

Bash is also not necessary.

vcaldas: Indeed. I had a complain about not finding these 2 when building locally. Will doublecheck the commands here

Docker/Dockerfile
vcaldas: That's quite interesting. Pairing to your message below, I am building it in the "wrong" way. Now, I build the container locally and then use on the composer. Really appreciate the feedback!

PR #3944, opened 2022-03-14
Update local volumes name to match readme

vraptor75011:


vraptor75011:
I did the install like this on two servers and it s ok.
If you want to keep the original version then an update to the docker readme may be necessary at the steps of directories creation starting with : mkdir -p Platform before creating all the Myxxxx directories

PR #3955, opened 2022-03-21
Allow 'chart' to be used at announcement conditions ; Make announcement node config show options as per documentation

PLuv:


ssplatt (approved):
👍

PR #3974, opened 2022-03-29
Contribution: Fixed Binance config

9808us:


9808us:
Binance works out of the box. No need of extra config

gitforkgit:
@Luis-Fernando-Molina can you maybe confirm it? I will merge anyway, but if there occur problems we will need to fix it.

PR #4006, opened 2022-04-13
Contribution: Education-Translations (Turkish) - Trading Logic

tigers40:


gitforkgit:
Is it a mistake that you delete so many lines?

PR #4014, opened 2022-04-19
Fixes and Adds. Will write a comment on GitHub

9808us:
This PR involves 3 files:

1) HistoricOHLCVs.js -> sandBox + many more parameters we can now use when connecting to exchanges. Now we can use "options" as expected (it was not correctly implemented) and set limit, rateLimit ecc whitout the needing of using "API" that was supposed to be used for fetching candles in an alternative way.

2) ExchangeAPI.js -> sandBox is now correctly working and also all the above parameters are now included so to match the initial instantiation with the exchange. Inside this file the function createOrder is now able to accept parameters as per ccxt manual.
Added 2 more parameters that can be configured at the exchange node and at position node (ie for reduceOnly for example)

3) OrdersCalculation.js -> Initial changes to the code so now SA can correctly trade and keep track of inverse markets like BTC/USDT:BTC

If PR is accepted I will then document the use of the new parameters at the Exchange and Order nodes

### DIFF
--- a/Projects/Algorithmic-Trading/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/APIs/ExchangeAPI.js
+++ b/Projects/Algorithmic-Trading/TS/Bot-Modules/Trading-Bot/Low-Frequency-Trading/APIs/ExchangeAPI.js
@@ -10,31 +10,58 @@
         finalize: finalize
     };
 
-    let tradingSystem
-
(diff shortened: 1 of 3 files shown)


9808us:
Followings are screenshot that can help understand.

Parameters set at the Market node:
image

More structured parameters set under Exchange node:
image

Parameters set for single order like a reduceOnly parameter:
image

Luis-Fernando-Molina:
Great job! please document and keep an eye at the issues / chats to see if someone has any issues related to this.

PR #4095, opened 2022-05-07
Contribution: German translation Foundation Nodes/Topics > LAN-Network

lindersi:


lindersi:
Again and again charset issues with my commits... And every time huge work to correct...

PR #4121, opened 2022-05-17
Added continue after error and warning of bad lines when reading CSV …

DJ05GaZ:
…file.


Luis-Fernando-Molina:
Can you please elaborate what was the problem and solution ?

PR #4157, opened 2022-05-28
Contribution: Improved Last Contribution, Loads & Checks Last File Date

theblockchainarborist:
The problem was the difference between local time and UNIX time. When the two times pointed to different days all data from that day would be deleted and replaced by only 1 API response each loop until the dates matched up again. I improved on the last code I submitted to fix this same issue by better defining the date of the last file on record. 
Tested during problem hours, and everything now functions as it should.


Luis-Fernando-Molina:
Awasome!

PR #4180, opened 2022-06-08
Improvement: Check for Updates at LAN Network Node

Smidy13:


Smidy13:
As I am lazy I dislike having to go to a different webpage to update each of my network nodes (when you have multiple). Also not knowing if I have restarted them or not annoys me.

This commit brings the option under the LAN Network Node for 'Check for Updates' which will check and update the corresponding node.

image

Also if a node has been update but the platform not restarted it will now show beneath the node.

image

Also there is a bug fix as the Web Socket connection would check repeatedly as the "lastTryToReconnectDatetime" variable wasn't being updated in the timeout IF statement (so would always be old).

PR #4208, opened 2022-06-27
Dashboard init.

devosonder:


harrellbm:
@devosonder This looks great! Just need to fix the one issue getting flagged by deep scan

PR #4229, opened 2022-07-24
Governance: Consolidation of Liquidity Rewards Pools into one

BastianMuc:
This change will introduce the harmonization of Liquidity Pools. By implementing this change, the distribution of rewards will become more fair, as it will no longer be possible to drive rewards for small liquidity pools by voting.

Going forward, liquidity will be rewarded identically, regardless what the paired asset or exchange is. The SA Token acts as the base value to harmonize liquidity across several different assets.

The PR also introduces initial structual enhancements to enable easier extension to other chains beyond BSC.

Note: Please merge the Governance-Plugins repo in timely proximity, as missing pools will otherwise lead to breaking changes.


ssplatt (approved):
i'm certainly not an expert in this area of the project but this looks ok to me

PR #4230, opened 2022-07-29
Contribution: System32 PATH is now converted to lowercase to avoid a rare bug when runing node setup

caoHenriques:


harrellbm (commented):
This looks like a great fix. Only thing with the Pull request is the strange deletions from the jquery dependencies.  @Luis-Fernando-Molina should jquery be tracked by this repo?

Luis-Fernando-Molina:
Probably not. Jquery was introduces when dragging was added to tutorials pages. I would try not to depend on that.

caoHenriques:
thank you
sorry I didn't noticed I also submited those 2 files, I will PR again but now make sure I PR only the file I fixed

PR #4232, opened 2022-07-31
Bitcoin-Factory Test Report Files

theblockchainarborist:
These are the Test Report Files from the Alpha and Quantum data mines.


BastianMuc:
Hi @theblockchainarborist , these reports will not be recognized for the governance process as they don't match the required filename pattern Testnet*.csv .

Do you want to change and re-upload?

PR #4233, opened 2022-08-01
Bitcoin Factory: Moving old reports to history folder for reduced parsing time

BastianMuc:
Also includes minor typo correction in Tutorial


BastianMuc:
... now also includes a fix at httpInterface.js which seems to remove the problem of failing app.contribute executions. Sorry for the mess in here, the rather slow merges during holiday period lead to messy PRs :)

PR #4237, opened 2022-08-03
Governance: Adding Multichain support for SA Token

BastianMuc:
This change introduces underlying multichain support to all areas of the Governance code working with blockchain balances and the SA Token (esp. User Balances, Reputation, Distribution), including the capability to communicate with the ETH mainnet and testnet.

This change does not contain any steps directly related to the deployment of new token editions (preparing these deployments is under preparation by Gary). It though establishes the fundamentals in the SA code for being able to handle several token editions on different chains.


Smidy13:
Will leave this one with Luis to review given it pertains to Governance.

Luis-Fernando-Molina:
ok, I am back home, will look at this after the current distribution happening soon.

Luis-Fernando-Molina:
Very good job!!!

PR #4258, opened 2022-08-13
Contribution: Updating some translation to investigate commit issues

BastianMuc:


BastianMuc:
PR now also contains the move of a malformed Bitcoin Factory Reports file to the History folder, so we get rid of the error messages when parsing

PR #4266, opened 2022-08-18
Fix for error introduced in last PR - Add check for if "undefined"

theblockchainarborist:


theblockchainarborist:
Looks like another odd ball error to take care of still.

PR #4277, opened 2022-08-21
Fix for Network Node Multiple Test Servers Working

theblockchainarborist:
Fixed a error that was causing the network node to mix up what server it should send finished test case results to.


Luis-Fernando-Molina:
good job!

PR #4289, opened 2022-08-24
update readme & remove excluded files in .vscode

teehanming:
the excluded files in .vscodev made me take a long time to find the My-Workspaces folder just to know it was hidden in .vscode.


harrellbm (changes requested):
Not sure if you meant to change the package-lock?

Also has it been discussed adding all these files back to the vscode search? I don't want to mess up anyone's workflow

teehanming:
I'm not sure why the package-lock included, I will make new pull request

PR #4300, opened 2022-08-27
Charting Space Update

theblockchainarborist:
Thanks to 9808us this update allows for the user to control both the minimum amount of data that is plotted on the charts and the maximum amount of data that can be plotted when the user scrolls the chart.

These updates can be controlled from the Space Style node under the charting space just like the other charting space UI updates.

Default settings:
"minCursorSize": "10"
"maxCursorSize": "30"

Test Settings:
"minCursorSize": "300"
"maxCursorSize": "700"


gitforkgit (commented):
The docs file looks great :+1:  . The other file get's reviewed by the @Superalgos/foundations-project

PR #4348, opened 2022-09-24
Contribution: Trading Signasl Improved with better syncronization based on Receiving Bot Waiting for Signals in order to run

Luis-Fernando-Molina:


Luis-Fernando-Molina:
⚠️⚠️⚠️ Major improvement: Today I am going to post a new version of Trading Signals with an improvement in the Synchronization between the Signal Receivers and Emitters.

Before this new version, the Trading Bot of a receiver was running every one minute, and the signal is broadcasted whenever the trading bot emitting the signals finishes. The problem was that if lets say for instance that the receiver was always running at 10 seconds of the start of a minute and the emitter at 15 seconds, then the receiver will always be almost a minute behind.

Worse than that actually, since each of the two are running their own data mining, which they both most likely run desynchronized from each other, maybe the emitter is seeing candle #300, and at the same time the receiver #301, or viceversa. This desynchronization between all this make it impossible to predict when the receiver was going to process the last signal.

To solve this problem, I added this parameter to the Trading Bot Instance, that when true, it forces the bot to wait until a signal is received (online) in order to run the next loop, in a similar way to when a bot waits for another process to finish.

So when you add this parameter to your trading bot, it will effectively wait for the next signal to arrive and then run itself with the assurance that the data needed to process is already there. This has a notable improvement in synchronicity, you will notice that the receiver bot runs 5 seconds after the emitter finished.

If only the trading bot could have that setting, we could still be in the situation that the instance of Superalgos receiving signals might be one candle behind at the Data Mining level. To solve this last synchronization problem, I also allowed the Sensor Bot Instance to run with the same parameter which would force it to ran and fetch the next candle every time it receives a signal, instead of doing it every minute.

So the perfect configuration of a Receiving Workspace should have the 
    "waitForSignalsToRunNextLoop": true
setting not at the Trading Bot Instance, but at the Sensor Bot Instance and the Trading Bot should be waiting for the Indicator Bot that build all candles to finish in order to run.

Configured in this way, we would have the optimal synchrony and between a set of receiving bots, the one who would process the signal first would be the one with most SA Balance at the User Profile.

This latest improvements essentially finishes the current version of Trading Signals, and with this we are ready to start trading in groups following the rule of the Ranking of the User Profiles by SA Tokens. 
⚠️⚠️⚠️

PR #4365, opened 2022-10-01
Bitcoin Factory: Archiving old Test Reports, fixes to new ones

BastianMuc:


BastianMuc:
This PR is handling a number of issues regarding Bitcoin Factory Governance Rewards:

1) Old rewards files from previous cycles are moved to Archive to reduce parsing time.
2) For the current reporting period, several files were uploaded for the same test run (i.e. containing duplicate records, leading to duplicate rewards). Files only containing duplicate records were deleted.
3) For yet unknown reasons, a rewards file had a broken structure (i.e. values are put in changing different columns within the file, making it unparseable). While we are investigating the root causes, I have manually created a file with a consistent structure and moved the original, faulty one to archive.

PR #4410, opened 2022-10-18
P2P Network : new P2P Network Node Reference node

vraptor75011:
Previous PR introduced the possibility to connect to a specific user P2p Network Node using by adding this key to the P2p Network Client Node : 
```json
"onlyConnectToNetworkNodeFromUserProfile": "vraptor75011"
```
Capture d’écran 2022-10-18 à 12 16 05

It is now replaced with the implementation of a node called P2p Network Node Reference to be added to the P2p Network Client Node and referencing the user profile P2p Network Node : 

Capture d’écran 2022-10-18 à 12 13 33


Luis-Fernando-Molina:
Good job!

PR #4426, opened 2022-10-22
Contribution: 

efrael:
Italian translation:

26 Files edited from the Tutorial Chapter
10 full Translations
16 edited outdated Translations
1 File translated from Concepts


gitforkgit (commented):
Everything looks fine. Someone from @Superalgos/foundations-project could look after the last file because I don't know about what it means (Superalgos    file).

PR #4443, opened 2022-10-25
Upgrade electron works with npm install

superharoun:


superharoun:
Capture d’écran 2022-10-25 à 23 31 45
Had this problem it doesn't exist any more

PR #4470, opened 2022-10-31
Feature/add submodule file

Awhiteweb:
Added .gotmodule file with plugin modules to allow docker build to run


Luis-Fernando-Molina:
hey! have you tested that app.update and the scripts to setup the plugins will run as usual after this change? I guess they would but would like to be sure.

Luis-Fernando-Molina:
Also please check that contributing ussing app.contribute will work, and via the contributions tab.

PR #4523, opened 2022-11-24
Feature/exporting docs

Awhiteweb:
New CLI process to export documentation as static website.

Includes updates to github workflows to reduce forked repos from trying to run all the actions.

Minor update to main UI app where icons function has been moved into new file. I have checked the app and this is all still working.

New export-docs and Exports directory containing process files and readme to help users understand how to run the command.

New dependencies added to package.json -> jsdom: for processing html files & yargs: for creating cleaner cli for running process.


Awhiteweb:
I can 'hack' the last deepscan issue if it is required to progress, but the error is because I have 2 placeholders that are string replaced during the build process

PR #4629, opened 2022-12-29
added branch setup matching for plugins

Awhiteweb:


Awhiteweb:
Deep scan results not relevant to this PR. Could also be PR's into master if wanting to apply fix.

PR #4634, opened 2023-01-01
Feature/credentials extraction

Awhiteweb:
Added a means to extract credentials from My-Workspaces so workspaces can be stored, shared and reused without needing to check for leaked credentials.

Secret values are removed from the workspace json before saving, keys/names are still stored and used as a reference to match with api values.

A new file is created in ./My-Secrets/workspace-credentials.json which uses a generic array of objects to store values:

```json
[
  { 
    "type": "github"|"exchange",
    "key: "api_key"
    "value: "api_secret"
  }
]
```

This also now means that credentials can be shared across workspaces more easily and it will mean if you need to update and API key, for example, your github token it will get updated for all workspaces. One concern is if you have many API secrets for a single key as could be the case for github.


Luis-Fernando-Molina:
Very good job!

PR #4771, opened 2023-02-27
added Vuetify framework to vue

end-debug:


NAmorim:
Can you clarify what is the goal here? This introduces a severe regression in the app.

end-debug:
> Can you clarify what is the goal here? This introduces a severe regression in the app.

to make building the app simpler for non-dev, like myself easier..

PR #4774, opened 2023-02-27
Trading Signals - New logic to follow bots with access control by token power

BastianMuc:
This PR introduces a large amount of changes to the P2P Networking infrastructure. After this change:
- Users can add a new node type "Followed Bot Reference" to their Social Trading Bots. This node type can be referenced to other's Social Trading Bots to signal followership.
- Network Nodes can now send Signals in a targeted way only to followers of a specific signal, no longer to everyone connected to the Network Node.
- Signal broadcasters can now define a minimum token power requirement at Social Trading Bots to follow their signals. Signals are only distributed to followers when they have allocated more than the requested token power to their Followed Bot Reference Nodes.
- Includes numerous changes to handle followers on all network levels.

This change is non-breaking (well, hopefully). When merged, users not defining a followership will still receive all signals from network nodes from now - the node will though put followers with allocated token power to the front of the queue. Should be ready for initial user testing.


Luis-Fernando-Molina:
Good job!

PR #4786, opened 2023-03-06
Contribution: Added "fixed" in addition to "percentage" - Merge POST distribution

cozed-gh:


BastianMuc:
Hey @cozed-gh , just saw this randomly - Please note that due to the introduction of Token Power for signals, there's now a second function managing Token Power on SA level located under Governance/SA/Function-Libraries/ProfileTokenPower.js .

For now, changes made to the Token Power logic need to be replicated there for consistency.

Awhiteweb:
There is a lot of duplication, it would be best to abstract the percentage function into a Utility class and pass in the node properties not just the node object.

cozed-gh:
> Hey @cozed-gh , just saw this randomly - Please note that due to the introduction of Token Power for signals, there's now a second function managing Token Power on SA level located under Governance/SA/Function-Libraries/ProfileTokenPower.js .
> 
> 
> 
> For now, changes made to the Token Power logic need to be replicated there for consistency.

Alright! Good to know! I have a look in that ProfileTokenPower.js and see what I can find.

cozed-gh:
> There is a lot of duplication, it would be best to abstract the percentage function into a Utility class and pass in the node properties not just the node object.

Yes I noticed this too. At of a start I'd just be happy to get it working. I see the 'power' from parent node is named differently from program to program. But it sure makes sense to pass it to the function instead.
Good point!

cozed-gh:
Will close this one and post another with somewhat more streamlined code.

- Function inside Utilities
- All nodes will have node.payload.tokenPower which allows the function to calculate only sending 'childNode' to it

PR #4789, opened 2023-03-10
Contribution: Added "amount" as option to "percentage" - Do not merge until after distribution

cozed-gh:


BastianMuc:
Hey @cozed-gh , I'm throwing in comments from the sideline again (sorry for that! :-) )
Just from looking at the code (not from testing), the implementation within ProfileTokenPower.js should normally not work, i.e. lead to a crash when executed on a network node. The background is the network node operates under a Task Server, from where access to any kind of UI.* function is normally not possible. If I understand things right, the utility function called would need to live in the SA.* space.

PR #4803, opened 2023-03-16
[SocialTrading] Moving to new logging framework

NAmorim:


NAmorim:
Appears some tests slipped into the PR. Closing.

PR #4810, opened 2023-03-18
Trading Signals: Fixes and Feedback from User Testing

BastianMuc:


Awhiteweb:
Can you abstract out the `parseFloat(tokenPower.toFixed(0)).toLocaleString('en')` into the SA.projects.governance.utilities.balances module as you are repeating it a lot on the code and would easier to manage it in the a module, then you could just have a function `toTokenPowerString(tokenPower)` that does the parsing, toFixed and setting the locale.

BastianMuc:
> Can you abstract out the `parseFloat(tokenPower.toFixed(0)).toLocaleString('en')` into the SA.projects.governance.utilities.balances module as you are repeating it a lot on the code and would easier to manage it in the a module, then you could just have a function `toTokenPowerString(tokenPower)` that does the parsing, toFixed and setting the locale.

Beautified accordingly, thanks for the feedback!

PR #4841, opened 2023-04-14
Revert "Event Server: Hardening Websockets Connections"

BastianMuc:
Reverts Superalgos/Superalgos#4833


BastianMuc:
Merging Revert as Hotfix.

PR #4859, opened 2023-04-30
Trading Signals: Deactivating signal broadcasting for non-followers

BastianMuc:
This change completes the transition to the requirement of allocating Token Power for following signals. Users not having established a follower relationship by adding a "Followed Bot Reference" node will be cut off from signal receipt.


Projects/Trading-Signals/TS/Modules/OutgoingTradingSignals.js
Awhiteweb, line 1: @BastianMuc you have a require statement on line 1, is it actually required?

BastianMuc: It actually is not. VSCode seems to have done some magic again here. Removed.
Thanks for watching!

PR #4865, opened 2023-05-03
Initial implemented database scanner 

harrellbm:


Awhiteweb:
Added some preliminary comments about missed console logs.

Also think we can abstract out the database layer so it is not bound to SQLite, but need to review on computer (not my phone) first.

Awhiteweb:
Where does the database data come from, is it another part of the system or a thrid party database?

harrellbm:
@Awhiteweb Right now it is set up to scan external third party data. I'm away from the comp for a short trip but I can message you a demo workspace and db file to test with. Once things are locked down then we can add a new native workspace

Awhiteweb:
OK, I think for now then it's fine to merge in after you remove/edit the vscode files. Then we can talk about migrating the persistence layer out of the network and into the foundations project, then add in an SQLite layer so the scanner is dependent on the particular db.

I can probably get this done in June and we can liaise on upgrading the scanner file

PR #4880, opened 2023-05-12
Fixed SidePanel Workspace, changed Sample to Native Workspaces

MarkTradeink:
Replacing original Label "Plugin Workspace" (now Sample) by final label "Native Workspace in all Documents, Top Menu and Side Panel Tab for Workspaces. 

Also fixed SidePanel Workspace loading issue


harrellbm:
@MarkTradeink This looks great! One note I would make is that the name of some json files still has sample in it rather than native. That's the only thing I noticed!

MarkTradeink:
@harrellbm Thank you for noticing! I forgot that point😅 I will add a new commit later

theblockchainarborist:
Nice work here!

harrellbm (approved):

PR #4882, opened 2023-05-12
Reverting yesterdays change to network node

theblockchainarborist:
Reverting from change that makes network node reset connections every 1hr.


theblockchainarborist:
I'm merging my own PR strictly because I'm removing a change from a PR I made yesterday.

PR #4883, opened 2023-05-12
removed left panels from view

Awhiteweb:
Commented out code that created the workspaces and tutorials tabs on the left of the canvas.

Tutorials will need updating to reflect these changes.


Awhiteweb:
@julian-molina The getting started tutorial will need updating when this gets merged into develop as it references the 'left hand workspace menu'. Not sure if it is referenced in any other tutorials.

harrellbm (approved):

harrellbm:
@Awhiteweb Looks like a small deep scan flag on an unused variable now

Luis-Fernando-Molina:
just seen this, one question: if there is no Workspaces left pannel, what happens when a workspace is loaded and does not have the node that enables the Workspaces menu? In those cases the Workspaces left panel would be needed unless something else was implemented.

Awhiteweb:
I have been looking at updating all the plugin workspaces so they have the workspace node. But it is also an option that we have that menu item as a permanent feature in the UI so it is not needed in the workspace json file.

harrellbm:
@Awhiteweb Personally I think it makes sense as a permanent feature with the ability to add/adjust it using the workspace nodes

Awhiteweb:
I have now added the workspace project as a permanent first menu item. I have also added specific ordering the initial project items - those listed in the ProjectsMenu.json - and a secondary ordering to all sub menu items. The submenu items with children take preference in alphabetical order followed by items with no children in alphabetical order 
Ordering example

harrellbm:
I'm sure this is outside the scope of this PR but I have noticed that very long lists of workspaces overflow the bottom of the screen without anyway to scroll. A needed improvement for the future

Awhiteweb:
Yes I am aware of this issue. I might look at it this month if I get a chance. I doubt this PR will go through by then as it requires the tutorials to be updated first.

Awhiteweb:
Closing as includes different work and will re-open with only required changes

PR #4924, opened 2023-06-24
Early Reinforcement Learning results. 

markeone:
Just a small test on the HannWindow strategy running some early tests.
Results are over 01-01-2023 to 23-06-2023. 
Ultimately this shows RL is back working. Will hope to make a public network and get more tests completed.
Forecast case 2 results


Awhiteweb:
@markeone can you merge in the most recent develop and see if this resolves the deepscan failure.

PR #4947, opened 2023-07-22
Governance: Adding zkSync as distribution option for SA Tokens

BastianMuc:
This PR introduces the required changes for enabling SA Token Distribution on the zkSync L2 scaling solution based on ETH. 

The PR establishes the required groundwork, but does not affect the actual distribution yet (ETH L1 remains configured as the main distribution channel until the community decides otherwise).

The change includes logic to correctly reflect reputation earned during zkSync distributions (rather high complexity, as zkSync is not using an Etherscan fork).


Platform/Client/web3Server.js
```diff
@@ -88,19 +88,23 @@ exports.newWeb3Server = function newWeb3Server() {
...
..., "inputs": [{ "indexed": true, "name": "burner", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Burn", "type": "event" }, { "anonymous": false, "inputs": [], "name": "Pause", "type": "event" }, { "anonymous": false, "inputs": [], "name": "Unpause", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "previousOwner", "type": "address" }, { "indexed": true, "name": "newOwner", "type": "address" }], "name": "OwnershipTransferred", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "blackListed", "type": "address" }, { "indexed": false, "name": "value", "type": "bool" }], "name": "Blacklist", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "owner", "type": "address" }, { "indexed": true, "name": "spender", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Approval", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "from", "type": "address" }, { "indexed": true, "name": "to", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Transfer", "type": "event" }]
```
Awhiteweb, line 91: You can move this value out into the parent object as it is repeated later in the code. There is also no point creating the object as a string and then immediately replacing it with an array.

Projects/Network/SA/Modules/AppBootstrapingProcess.js
```diff
@@ -421,21 +421,20 @@ exports.newNetworkModulesAppBootstrapingProcess = function newNetworkModulesAppB
...
..., "inputs": [{ "indexed": true, "name": "burner", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Burn", "type": "event" }, { "anonymous": false, "inputs": [], "name": "Pause", "type": "event" }, { "anonymous": false, "inputs": [], "name": "Unpause", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "previousOwner", "type": "address" }, { "indexed": true, "name": "newOwner", "type": "address" }], "name": "OwnershipTransferred", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "blackListed", "type": "address" }, { "indexed": false, "name": "value", "type": "bool" }], "name": "Blacklist", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "owner", "type": "address" }, { "indexed": true, "name": "spender", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Approval", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "from", "type": "address" }, { "indexed": true, "name": "to", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Transfer", "type": "event" }]
```
Awhiteweb, line 424: assign the ABI straight away instead of using a string and then replacing with an object.

Projects/Network/SA/Modules/ProfileManagerAppBootstrapingProcess.js
```diff
@@ -384,21 +384,20 @@ exports.newNetworkModulesProfileManagerAppBootstrapingProcess = function newNetw
...
..., "inputs": [{ "indexed": true, "name": "burner", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Burn", "type": "event" }, { "anonymous": false, "inputs": [], "name": "Pause", "type": "event" }, { "anonymous": false, "inputs": [], "name": "Unpause", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "previousOwner", "type": "address" }, { "indexed": true, "name": "newOwner", "type": "address" }], "name": "OwnershipTransferred", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "blackListed", "type": "address" }, { "indexed": false, "name": "value", "type": "bool" }], "name": "Blacklist", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "owner", "type": "address" }, { "indexed": true, "name": "spender", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Approval", "type": "event" }, { "anonymous": false, "inputs": [{ "indexed": true, "name": "from", "type": "address" }, { "indexed": true, "name": "to", "type": "address" }, { "indexed": false, "name": "value", "type": "uint256" }], "name": "Transfer", "type": "event" }]
```
Awhiteweb, line 387: same as previous comments

Awhiteweb:
As this ABI is used in multiple files, it would make sense that it lives as a constant value in a global file that they can all reference.

BastianMuc:
Some code cleaning done. Will handle the global file suggestion at a later point, as I'm currently lacking a suitable spot on SA level (globals for Governance live at UI, moving that will be a slightly larger exercise).

PR #4985, opened 2023-09-19
Feature/system translations

Awhiteweb:
Adding ability to add translations to the system text. Has required some small changes to some of the styles but will need some continual development as translations are added to make sure text is visible with each language.


theblockchainarborist:
I have reviewed the code and everything looks solid. Merging because it has sat here awhile. Nice work!!

PR #5014, opened 2023-11-08
Addition of Missing Asset Logo

prepheadrus:


Projects/Foundations/Schemas/App-Schema/asset.json
BastianMuc, line 1: Just coincidentally saw that RPL is included here twice with two different logo names - could you briefly check for duplicates? Thanks!

BastianMuc (changes requested):

Projects/Foundations/Schemas/App-Schema/asset.json
prepheadrus: Ohh.. Yes, you're right. I'll fix it...

PR #5038, opened 2023-12-21
Feature/goverance csv export

Awhiteweb:
Add ability to generate a CSV governance report from the UI Governance Tab. This will download a full report of the current state of governance.


Awhiteweb:
The majority of the file changes are the addition of 'exports' to the UI functions so they can be used later as a console generator.

PR #5127, opened 2024-04-16

cozed-gh:


cozed-gh:
Missed to clean out some garbage code.

PR #5146, opened 2024-09-05

AngloMnemonic1:


BastianMuc (changes requested):
This Pull Request would also delete an image file (study-bot.png) - I assume this is not intended. Can you re-submit the PR without this file deletion?

PR #5192, opened 2025-03-03
Created Modules

end-debug:
Added a Vuex module for handling persons, posts and persist socialEntities using localStorage
Keeps post-related state separate for better maintainability
Improves scalability as more features are added


Social-Trading/Vue-UI/src/store/modules/personas.js
BastianMuc, line 1: Is line 185 really intending to say:
followersResponse.data.data === '[}'

Never seen this, so just noting as to me it looks like a kind-of-bracket-typo...

end-debug: I didn't see that and yes,  it's a typo

BastianMuc: Let me know if you want to submit a correction, or if I shall merge as-is :)

end-debug: i submitted a fix

